PR #3, opened 2019-12-05
Feature/apps 1588 create studies page

anzioka:
Created page to display list of studies. At the top of the page there is a 'Create Study' button will show modal with form to create a new page (not yet implemented) Each study is rendered in a LUK card and contains basic info about the study. The number label at the top right of the page is a placeholder for the number of participants and will be replaced by the actual number.  On clicking each card details about the study are displayed (haven't yet implemented this). 

study_page


UnsungHero97 (changes requested):
Please run `npm run lint` and `npm run flow` and fix any errors.

src/utils/constants/DataModelConstants.js
```diff
@@ -0,0 +1,58 @@
...
+  USER: 'nc.SystemUserName',
+  LATITUDE: 'location.latitude',
+  LONGITUDE: 'location.longitude',
+  ALTITUDE: 'ol.altitude',
+
+  // other
+  STATUS: 'ol.status',
+  TIMESTAMP: 'date.completeddatetime',
+  UUID: 'general.stringid'
+};
```
giuliacampana: Looks good! One thing to note is that we usually store property types (and entity types) as `FullyQualifiedNames`. So, for example: 

```
import { Models } from 'lattice';

const { FullyQualifiedName } = Models;

export const PROPERTY_TYPES = {
  PERSON_ID: new FullyQualifiedName('nc.SubjectIdentification'),
  PARTICIPANT_ID: new FullyQualifiedName('justice.xref'),
  ...
```

This is what I'm referring to: https://github.com/openlattice/lattice-js/blob/develop/src/models/FullyQualifiedName.js.

src/utils/constants/DataModelConstants.js
```diff
@@ -0,0 +1,58 @@
+export const ENTITY_SETS = {
```
UnsungHero97: This belongs in `core/edm/constants/EntitySetNames.js`.

src/utils/constants/DataModelConstants.js
```diff
@@ -0,0 +1,58 @@
...
+  PARTICIPANTS_PREFIX: 'chronicle_participants_',
+  PREPROCESSED_DATA: 'chronicle_preprocessed_app_data',
+
+  // association entity sets
+  RECORDED_BY: 'chronicle_recorded_by',
+  USED_BY: 'chronicle_used_by',
+  PARTICIPATED_IN: 'chronicle_participated_in'
+};
+
+export const PROPERTY_TYPES = {
```
UnsungHero97: This belongs in `core/edm/constants/FullyQualifiedNames.js`.

src/utils/constants/DataModelConstants.js
```diff
@@ -0,0 +1,58 @@
...
+  LONGITUDE: 'location.longitude',
+  ALTITUDE: 'ol.altitude',
+
+  // other
+  STATUS: 'ol.status',
+  TIMESTAMP: 'date.completeddatetime',
+  UUID: 'general.stringid'
+};
+
+export const GENERAL_STRING_ID_PT_ID = 'ee3a7573-aa70-4afb-814d-3fad27cda988';
```
UnsungHero97: What is this id?

UnsungHero97 (changes requested):

src/components/StudyCard.js
```diff
@@ -0,0 +1,101 @@
...
+  handleCardClick : PropTypes.func;
+}
+
+//  TODO : get the number of participants for a study;
+class StudyCard extends Component<Props> {
+  render() {
+    const { study, handleCardClick } = this.props;
+    const numParticipants = 3; // TODO: change this to the actual number of participants
+    return (
+      <Card onClick={(event) => handleCardClick(event)} data-study-id={study.id}>
```
giuliacampana: Not sure if the linter should have caught this, but typically, for prop names, we use `camelCase` as opposed to `kebab-case` (is that what it's called?), as you have with the `onClick` prop:
```
dataStudyId={study.Id}

File: src/containers/studies/StudiesContainer.js [hunk 6]
```diff
@@ -0,0 +1,130 @@
...
+            ? (
+              <CenterText>
+              Sorry, no studies were found. Please try refreshing the page, or contact support.
+              </CenterText>
+            )
+            : (
+              <CardGrid>
+                {formatedStudies.map((study) => (
+                  <StudyCard key={study.id} study={study} handleCardClick={this.handleCardClick} />
+                ))}
```
giuliacampana: Also wondering if the linter should have caught this... but the brackets should each be on their own line! Also might be a good idea in this case, with 3 props, to put them each on their own line:
```
<CardGrid>
  {
    formatedStudies.map((study) => (
      <StudyCard
          key={study.id}
          study={study}
          handleCardClick={this.handleCardClick} />
    ))
  }
</CardGrid>

File: src/containers/studies/StudiesSagas.js [hunk 7]
```diff
@@ -0,0 +1,127 @@
...
+      searchEntitySetDataWorker,
+      searchEntitySetData({
+        entitySetId,
+        searchOptions: {
+          maxHits: 1000,
+          searchTerm: '*',
+          start: 0,
+        },
+      })
+    );
```
giuliacampana: I'm just curious why the `searchEntitySetData` endpoint is used here, as opposed to simply using `DataApi.getEntitySetData`. I see it's used in the original `StudiesSagas` too — @UnsungHero97 I assume this endpoint was used to limit the number of results—but if there were 1000 studies, that still seems like a pretty large number of results to display... unless the intention was to paginate if over a certain threshold were returned? In that case, then I think it would make more sense to use the `executeSearch` endpoint, since now it's easy to get paginated results from the backend, ya?

src/utils/constants/DataModelConstants.js
anzioka: ported over the constants file from the old project without making any changes ... but I have removed the Id since there doesn't seem to be any use for it

src/components/StudyCard.js
UnsungHero97: @giuliacampana this is actually correct. `data-` attributes are special:

https://developer.mozilla.org/en-US/docs/Web/API/HTMLOrForeignElement/dataset

package.json
```diff
@@ -42,6 +42,7 @@
     "lodash": "~4.17.10",
     "loglevel": "~1.6.0",
     "polished": "~3.4.0",
+    "prop-types": "^15.7.2",
```
UnsungHero97: We use Flow instead of `prop-types`. Please remove.

https://flow.org/

package.json
```diff
@@ -103,7 +104,7 @@
     "flow": "flow",
     "lint": "npm-run-all lint:css lint:js",
     "lint:css": "stylelint './src/**/*.js'",
-    "lint:js": "eslint --config .eslintrc src",
+    "lint:js": "eslint --config .eslintrc --fix src",
```
UnsungHero97: I don't think autofixing is a good idea. Please remove.

src/components/StudyCard.js
```diff
@@ -0,0 +1,101 @@
...
+
+const StudyName = styled.h2`
+  font-size: 18px;
+  font-weight: normal;
+  margin: 0;
+  padding: 0;
+`;
+/* stylelint-disable value-no-vendor-prefix, property-no-vendor-prefix */
+const StudyDescription = styled.p`
+  font-size: 14px;
```
UnsungHero97: Please sort in alphabetical order.

src/components/StudyCard.js
```diff
@@ -0,0 +1,101 @@
...
+`;
+
+const ParticipantsIcon = styled(FontAwesomeIcon).attrs({
+  icon: faUsers
+})`
+  margin-right: 10px;
+  color: ${NEUTRALS[1]};
+`;
+
+type Props = {
```
UnsungHero97: I'm pretty sure Flow will complain about this. You'll want to do the following:

```
type Props = {
  handleOnClickCard :() => void; // you can also do :Function
  study :Object;
};
```

src/components/StudyCard.js
UnsungHero97: `onClick={handleCardClick}`

src/containers/app/AppContainer.js
```diff
@@ -113,13 +114,12 @@ class AppContainer extends Component<Props> {
       <AppContainerWrapper>
         <AppHeaderWrapper appIcon={OpenLatticeIcon} appTitle="Chronicle" logout={this.logout} user={user}>
           <AppNavigationWrapper>
-            <NavLink to={Routes.ROOT} />
-            <NavLink to={Routes.HOME}>Home</NavLink>
-            <NavLink to="/tab1">Tab 1</NavLink>
-            <NavLink to="/tab2">Tab 2</NavLink>
+            <NavLink to={Routes.STUDIES} />
```
UnsungHero97, line 117: You'll want another `NavLink` so that it shows as a tab:

```
<NavLink to={Routes.STUDIES} />
<NavLink to={Routes.STUDIES}>Studies</NavLink>
```

src/containers/studies/StudiesActions.js
```diff
@@ -0,0 +1,16 @@
+
```
UnsungHero97, line 4: You need this at the top:

```
/*
 * @flow
 */
```

src/containers/studies/StudiesContainer.js
```diff
@@ -0,0 +1,130 @@
...
+}
+const mapStateToProps = (state) => ({
+  studiesReqState: state.getIn(['studies', GET_STUDIES, 'requestState']),
+  studies: state.getIn(['studies', 'studies'], List())
+});
+
+const mapDispatchToProps = (dispatch) => ({
+  actions: bindActionCreators({
+    goToRoute: RoutingActions.goToRoute,
+    getStudies,
```
UnsungHero97, line 126: Please sort alphabetically.

src/containers/studies/StudiesContainer.js
```diff
@@ -0,0 +1,130 @@
...
+class StudiesContainer extends Component<Props> {
+  componentDidMount() {
+    const { actions } = this.props;
+    actions.getStudies();
+  }
+  openCreateStudyModal = () => {
+    // TODO: open modal to create study
+  }
+
+  formatStudy = (study: Map) => ({
```
UnsungHero97: I don't think this is necessary.

src/containers/studies/StudiesContainer.js
```diff
@@ -0,0 +1,130 @@
...
+      <>
+        <ContainerHeader>
+          <h1> Studies </h1>
+          <Button mode="primary" onClick={this.openCreateStudyModal}> Create Study </Button>
+        </ContainerHeader>
+        {
+          formatedStudies.isEmpty()
+            ? (
+              <CenterText>
+              Sorry, no studies were found. Please try refreshing the page, or contact support.
```
UnsungHero97: Indent.

src/containers/studies/StudiesContainer.js
```diff
@@ -0,0 +1,130 @@
...
+              </CardGrid>
+            )
+        }
+        {this.openCreateStudyModal()}
+      </>
+    );
+  }
+}
+const mapStateToProps = (state) => ({
+  studiesReqState: state.getIn(['studies', GET_STUDIES, 'requestState']),
```
UnsungHero97: Please use the following pattern:

```
[GET_STUDIES]: state.getIn(['studies', GET_STUDIES, 'requestState'])
// or...
requestStates: {
  [GET_STUDIES]: state.getIn(['studies', GET_STUDIES, 'requestState'])
}
```

src/containers/studies/StudiesSagas.js
```diff
@@ -0,0 +1,127 @@
...
+    yield put(getStudies.finally(action.id));
+  }
+}
+
+function* getStudiesWatcher() {
+  yield takeEvery(GET_STUDIES, getStudiesWorker);
+}
+
+export {
+  getStudiesWorker,
```
UnsungHero97, line 55: Please sort alphabetically.

src/containers/studies/StudiesSagas.js
```diff
@@ -0,0 +1,127 @@
...
+    const entitySetName = `${ENTITY_SETS.PARTICIPANTS_PREFIX}${studyId}`;
+    return yield call(EntitySetsApi.getEntitySetId, entitySetName);
+  }
+  catch (error) {
+
+    return null;
+  }
+
+}
+function* getStudiesPermissionsWorker(action: SequenceAction) {
```
UnsungHero97: I don't think any of the `getStudiesPermissions`-related things need to exist. None of it is used in the old app. Please remove.

src/containers/studies/StudiesSagas.js
```diff
@@ -0,0 +1,127 @@
...
+    yield put(getStudiesPermissions.failure(action.id, error));
+  }
+  finally {
+    yield put(getStudiesPermissions.finally(action.id));
+  }
+
+  return response;
+}
+
+function* getStudiesPermissionsWatcher() {
```
UnsungHero97: Please remove.

src/containers/studies/StudiesActions.js
```diff
@@ -0,0 +1,16 @@
...
+import { newRequestSequence } from 'redux-reqseq';
+import type { RequestSequence } from 'redux-reqseq';
+
+const GET_STUDIES :'GET_STUDIES' = 'GET_STUDIES';
+const getStudies :RequestSequence = newRequestSequence(GET_STUDIES);
+
+const GET_STUDIES_PERMISSIONS: 'GET_STUDIES_PERMISSIONS' = 'GET_STUDIES_PERMISSIONS';
+const getStudiesPermissions: RequestSequence = newRequestSequence(GET_STUDIES_PERMISSIONS);
```
UnsungHero97: I don't think any of the `getStudiesPermissions`-related things need to exist. None of it is used in the old app. Please remove.

src/containers/studies/StudiesSagas.js
```diff
@@ -0,0 +1,127 @@
...
+  yield takeEvery(GET_STUDIES_PERMISSIONS, getStudiesPermissionsWorker);
+}
+
+function* getStudiesWorker(action :SequenceAction) {
+  try {
+    yield put(getStudies.request(action.id));
+    const entitySetId = yield call(EntitySetsApi.getEntitySetId, ENTITY_SETS.CHRONICLE_STUDIES);
```
UnsungHero97: Let's chat.

src/containers/studies/StudiesSagas.js
UnsungHero97: Yes, let's use `getEntitySetData` for now. We can get more fancy in the future.

UnsungHero97 (changes requested):

src/components/StudyCard.js
giuliacampana: Oops, sorry. My bad!

generateEntitySets.js
```diff
@@ -0,0 +1,48 @@
...
+          .setName(name)
+          .setTitle(name)
+          .build();
+      });
+      EntitySetsApi.createEntitySets(entitySets);
+    });
+};
+
+// uncomment to execute
+// generateEntitySets();
```
solomontang: These comments and the lack exports suggest this file does not need to be committed. Please remove.

src/components/StudyCard.js
solomontang: There are lots of nice plugins you can find for your IDE of choice that will do this for you.

src/components/StudyCard.js
```diff
@@ -0,0 +1,101 @@
...
+
+const { NEUTRALS } = Colors;
+
+const StudyName = styled.h2`
+  font-size: 18px;
+  font-weight: normal;
+  margin: 0;
+  padding: 0;
+`;
+/* stylelint-disable value-no-vendor-prefix, property-no-vendor-prefix */
```
solomontang: You should not have to apply this rule. I am using webkit prefixes just fine without any linting errors

src/components/StudyCard.js
solomontang: I like to call it skewer-case but kebabs are tasty

src/containers/studies/StudiesContainer.js
```diff
@@ -0,0 +1,130 @@
...
+  studies: List<*>;
+}
+class StudiesContainer extends Component<Props> {
+  componentDidMount() {
+    const { actions } = this.props;
+    actions.getStudies();
+  }
+  openCreateStudyModal = () => {
+    // TODO: open modal to create study
+  }
```
solomontang, line 72: Remove unused code

src/utils/constants/DataModelConstants.js
solomontang: Probably not good to hard code this value...

solomontang (changes requested):

.gitignore
```diff
@@ -4,6 +4,6 @@ coverage/
 npm-debug.log
 package-lock.json
 node_modules/
-
+generateEntitySets.js
```
solomontang, line 7: Please don't check in any changes to .gitignore if they're only meant for your local environment. You can imagine how this file would quickly become bloated if everyone did this. Please remove.

If you're using generateEntitySets to behaviorally test your components and functions, you might as well formalize the process and make it part of your testing suite instead of commenting it out.

package.json
solomontang: This is still not resolved

src/components/StudyCard.js
solomontang: turns out this might just be a problem with my ide. carry on

package.json
UnsungHero97: @anzioka please remove the `prop-types` dependency.

src/components/StudyCard.js
```diff
@@ -0,0 +1,108 @@
...
+  display: -webkit-box;
+  font-size: 14px;
+  font-weight: normal;
+  margin: 0;
+  overflow-wrap: break-word;
+  overflow: hidden;
+  padding: 0;
+  text-overflow: ellipsis;
+`;
+/* stylelint-disable */
```
UnsungHero97: This should be `stylelint-enable`.

src/components/StudyCard.js
```diff
@@ -0,0 +1,108 @@
...
+//  TODO : get the number of participants for a study;
+class StudyCard extends Component<Props> {
+  handleCardClick = () => {
+    // to implement this
+  }
+  render() {
+    const { study } = this.props;
+    const numParticipants = 3; // TODO: change this to the actual number of participants
+    return (
+      <Card onClick={this.handleCardClick} data-study-id={study.id}>
```
UnsungHero97: The type annotation shows that `study` is a `Map`, so this should be `study.get('id')`.

src/components/StudyCard.js
```diff
@@ -0,0 +1,108 @@
...
+        <CardHeader>
+          <StudySummary>
+            <StudyName>
+              {study.name}
```
UnsungHero97: The type annotation shows that study is a Map, so this should be `study.get('name')`.

src/components/StudyCard.js
```diff
@@ -0,0 +1,108 @@
...
+            <StudyParticipants>
+              <ParticipantsIcon />
+              {numParticipants}
+            </StudyParticipants>
+          </StudySummary>
+
+        </CardHeader>
+        <CardSegment>
+          <StudyDescription>
+            {study.description}
```
UnsungHero97: The type annotation shows that study is a Map, so this should be `study.get('description')`.

src/containers/studies/StudiesActions.js
```diff
@@ -0,0 +1,19 @@
...
+ * @flow
+ */
+
+import { newRequestSequence } from 'redux-reqseq';
+import type { RequestSequence } from 'redux-reqseq';
+
+const GET_STUDIES :'GET_STUDIES' = 'GET_STUDIES';
+const getStudies :RequestSequence = newRequestSequence(GET_STUDIES);
+
+const GET_STUDIES_PERMISSIONS :'GET_STUDIES_PERMISSIONS' = 'GET_STUDIES_PERMISSIONS';
```
UnsungHero97: `GET_STUDIES_PERMISSIONS` and `getStudiesPermissions` is not used anywhere. Please remove.

src/core/edm/EDMSagas.js
```diff
@@ -73,7 +80,36 @@ function* getEntityDataModelTypesWatcher() :Generator<*, *, *> {
...
+    );
+
+    if (response.error) {
+      throw response.error;
+    }
+
+    yield put(getAllEntitySetIds.success(action.id, response.data));
+  }
+  catch (error) {
+    yield put(getAllEntitySetIds.failure(action.id, error));
```
UnsungHero97, line 101: Add a log for the error:

```
  catch (error) {
    LOG.error(action.type, error);
    yield put(getAllEntitySetIds.failure(action.id, error));
  }
```

src/core/edm/EDMSagas.js
```diff
@@ -73,7 +80,36 @@ function* getEntityDataModelTypesWatcher() :Generator<*, *, *> {
...
+  }
+  finally {
+    yield put(getAllEntitySetIds.finally(action.id));
+  }
+
+}
```
UnsungHero97: `getAllEntitySetIdsWorker` is called in `initializeApplication`. It's good practice to have workers return something if they are directly called by other workers. Take a look at `getEntityDataModelTypesWorker` above and how it returns `workerResponse` at the end.

```
function* getAllEntitySetIdsWorker(action :SequenceAction) :Generator<*, *, *> {

  const workerResponse :Object = {};

  try {
    ...
  }
  catch (error) {
    LOG.error(action.type, error);
    workerResponse.error = error;
    yield put(getAllEntitySetIds.failure(action.id, error));
  }
  finally {
    ...
  }

 return workerResponse;
}
```

src/containers/studies/StudiesSagas.js
```diff
@@ -0,0 +1,53 @@
...
+    );
+    const response = yield call(getEntitySetDataWorker, getEntitySetData({ entitySetId }));
+
+    if (response.error) {
+      throw response.error;
+    }
+    yield put(getStudies.success(action.id, response.data));
+  }
+  catch (error) {
+    yield put(getStudies.failure(action.id, error));
```
UnsungHero97, line 42: Add a log for the error:

```
  catch (error) {
    LOG.error(action.type, error);
    yield put(getStudies.failure(action.id, error));
  }
```

src/containers/studies/StudiesReducer.js
```diff
@@ -0,0 +1,37 @@
...
+
+const INITIAL_STATE :Map<*, *> = fromJS({
+  [GET_STUDIES]: { requestState: RequestStates.STANDBY },
+  studies: List(),
+});
+
+export default function studiesReducer(state :Map<*, *> = INITIAL_STATE, action :Object) {
+  switch (action.type) {
+    case getStudies.case(action.type): {
+      const seqAction = action;
```
UnsungHero97:
```
import type { SequenceAction } from 'redux-reqseq';
...
const seqAction :SequenceAction = action;
```

src/containers/studies/StudiesContainer.js
```diff
@@ -0,0 +1,136 @@
...
+  studiesReqState: state.getIn(['studies', GET_STUDIES, 'requestState']),
+  studies: state.getIn(['studies', 'studies'], List())
+});
+
+const mapDispatchToProps = (dispatch :Function) => ({
+  actions: bindActionCreators({
+    goToRoute: RoutingActions.goToRoute,
+    getStudies,
+  }, dispatch)
+});
```
UnsungHero97: Here's how I order these:

```
{
  // first are shorthand properties, i.e. keys without values
  thing,
  // next are computed properties
  [THING]: state.get(THING),
  // last are regular properties
  thingies: state.get('thingies'),
};
```

src/containers/studies/StudiesContainer.js
UnsungHero97: @anzioka ping

src/containers/app/AppSagas.js
```diff
@@ -33,6 +35,7 @@ function* initializeApplicationWorker(action :SequenceAction) :Generator<*, *, *
     yield put(initializeApplication.request(action.id));
     const responses :Object[] = yield all([
       call(getEntityDataModelTypesWorker, getEntityDataModelTypes()),
+      call(getAllEntitySetIdsWorker, getAllEntitySetIds()),
```
UnsungHero97, line 38: You should fail the saga if `getAllEntitySetIds` fails.

src/containers/studies/StudiesContainer.js
```diff
@@ -0,0 +1,136 @@
...
+  }
+
+  formatStudy = (study :Map) => ({
+    id: study.getIn([PROPERTY_TYPE_FQNS.STUDY_ID, 0]),
+    name: study.getIn([PROPERTY_TYPE_FQNS.STUDY_NAME, 0]),
+    description: study.getIn([PROPERTY_TYPE_FQNS.STUDY_DESCRIPTION, 0]),
+    group: study.getIn([PROPERTY_TYPE_FQNS.STUDY_GROUP, 0]),
+    version: study.getIn([PROPERTY_TYPE_FQNS.STUDY_VERSION, 0]),
+    email: study.getIn([PROPERTY_TYPE_FQNS.STUDY_EMAIL, 0])
+  });
```
UnsungHero97: I don't think this is necessary for 2 reasons:
  1. you're effectively doing `study.toJS()`, which should be avoided
  2. doing this defeats the purpose of using `immutable.js`

I recommend removing `formatStudy` and using `study` directly as an immutable map.

UnsungHero97 (changes requested):

src/containers/app/AppSagas.js
anzioka: Modified to return an error in an object if saga fails

UnsungHero97 (approved):

src/containers/studies/StudiesContainer.js
solomontang: Moving forward, you should avoid checking in code for future features. Not necessary to resolve now. just thought i'd mention it.

src/containers/studies/components/StudyCard.js
```diff
@@ -0,0 +1,110 @@
...
+}
+
+//  TODO : get the number of participants for a study;
+class StudyCard extends Component<Props> {
+  handleCardClick = () => {
+    // to implement this
+  }
+  render() {
+    const { study } = this.props;
+    const numParticipants = 3; // TODO: change this to the actual number of participants
```
solomontang, line 85: No change necessary for this PR. But again, try not to check in code that is not intended for production

solomontang (approved):

PR #4, opened 2019-12-09
Feature/apps 1591 create study modal

anzioka:
Created an ActionModal for creating a new study. Functionality to create study will come later.

create_study


package.json
```diff
@@ -37,6 +37,7 @@
     "immutable": "4.0.0-rc.9",
     "lattice": "0.71.1",
     "lattice-auth": "0.16.0",
+    "lattice-fabricate": "^0.7.2",
```
UnsungHero97: Please use `~` instead of `^`.

src/containers/studies/StudiesActions.js
```diff
@@ -8,7 +8,12 @@ import type { RequestSequence } from 'redux-reqseq';
 const GET_STUDIES :'GET_STUDIES' = 'GET_STUDIES';
 const getStudies :RequestSequence = newRequestSequence(GET_STUDIES);
 
+const CREATE_STUDY :'CREATE_STUDY' = 'CREATE_STUDY';
+const createStudy :RequestSequence = newRequestSequence(CREATE_STUDY);
+
 export {
   GET_STUDIES,
-  getStudies
+  CREATE_STUDY,
```
UnsungHero97, line 15: Please sort.

src/containers/studies/StudiesContainer.js
```diff
@@ -62,17 +63,36 @@ type Props = {
   studies :List;
 };
 
-class StudiesContainer extends Component<Props> {
+type State = {
+  isCreateStudyModalVisible :boolean;
+}
```
UnsungHero97: You need a semi-colon here.

```
type State = {
  ...
}; <- semicolon
```

src/containers/studies/StudiesReducer.js
```diff
@@ -8,11 +8,13 @@ import type { SequenceAction } from 'redux-reqseq';
 
 import {
   GET_STUDIES,
+  CREATE_STUDY,
   getStudies
 } from './StudiesActions';
 
 const INITIAL_STATE :Map<*, *> = fromJS({
   [GET_STUDIES]: { requestState: RequestStates.STANDBY },
+  [CREATE_STUDY]: { requestState: RequestStates.STANDBY },
```
UnsungHero97, line 17: Please sort.

src/containers/studies/components/CreateStudyForm.js
```diff
@@ -0,0 +1,70 @@
...
+  version: {
+    classNames: 'column-span-12'
+  }
+};
+
+type Props = {
+  formData :Object;
+  handleOnChange :Function;
+  isSubmitting :boolean;
+}
```
UnsungHero97: Missing semicolon.

src/containers/studies/components/CreateStudyForm.js
```diff
@@ -0,0 +1,70 @@
...
+    },
+    version: {
+      title: 'Study Version',
+      type: 'string'
+    },
+    email: {
+      title: 'Contact Email',
+      type: 'string'
+    }
+  },
```
UnsungHero97: You should wrap this in another layer of `properties` because it'll render nicer in the modal. I'll show you.

src/containers/studies/components/CreateStudyModal.js
```diff
@@ -0,0 +1,111 @@
...
+  data :Object;
+  isSubmitting :boolean;
+}
+
+
+const ModalBodyWrapper = styled.div`
+  min-width: 440px;
+`;
+
+const initialFormDataState :Object = {
```
UnsungHero97: If you're pulling out initial state into a constant, why is it not being used anywhere?

src/containers/studies/components/CreateStudyModal.js
```diff
@@ -0,0 +1,111 @@
...
+  email: '',
+  group: '',
+  name: '',
+  version: '',
+};
+class CreateStudyModal extends Component<Props, State> {
+  constructor(props :Props) {
+    super(props);
+    this.state = {
+      data: { ...initialFormDataState },
```
UnsungHero97: You don't need to spread. `data: initialFormDataState` is identical. Also, I recommend removing `initialFormDataState` and just setting the state here, unless you plan on reusing `initialFormDataState` somewhere else.

src/containers/studies/components/CreateStudyModal.js
```diff
@@ -0,0 +1,111 @@
...
+  isVisible :boolean;
+  requestStates:{
+    CREATE_STUDY :RequestState
+  }
+};
+
+type State = {
+  data :Object;
+  isSubmitting :boolean;
+}
```
UnsungHero97: Missing semicolon.

src/containers/studies/components/CreateStudyModal.js
```diff
@@ -0,0 +1,111 @@
...
+  handleOnCloseModal :Function;
+  isVisible :boolean;
+  requestStates:{
+    CREATE_STUDY :RequestState
+  }
+};
+
+type State = {
+  data :Object;
+  isSubmitting :boolean;
```
UnsungHero97: We've moved away from using booleans to indicate request state. Please use a proper `RequestState`, which is probably `CREATE_STUDY`.

Separately, from what I can see, `isSubmitting` isn't real. It is not passed in from a parent component. Please double check this prop.

src/containers/studies/components/CreateStudyModal.js
```diff
@@ -0,0 +1,111 @@
...
+import CreateStudyForm from './CreateStudyForm';
+
+import { CREATE_STUDY } from '../StudiesActions';
+
+type Props = {
+  handleOnCloseModal :Function;
+  isVisible :boolean;
+  requestStates:{
+    CREATE_STUDY :RequestState
+  }
```
UnsungHero97: Missing semicolon.

src/containers/studies/components/CreateStudyModal.js
```diff
@@ -0,0 +1,111 @@
...
+import { connect } from 'react-redux';
+import { RequestStates } from 'redux-reqseq';
+import type { RequestState } from 'redux-reqseq';
+
+import CreateStudyForm from './CreateStudyForm';
+
+import { CREATE_STUDY } from '../StudiesActions';
+
+type Props = {
+  handleOnCloseModal :Function;
```
UnsungHero97: The `:Function` type should be avoided. The annotation for this should be `() => void;`.

https://flow.org/en/docs/types/functions/#toc-function-type

src/containers/studies/components/CreateStudyModal.js
```diff
@@ -0,0 +1,111 @@
...
+
+type State = {
+  data :Object;
+  isSubmitting :boolean;
+}
+
+
+const ModalBodyWrapper = styled.div`
+  min-width: 440px;
+`;
```
UnsungHero97, line 30: No change required.

This probably belongs in `src/components/modals`, similar to how `lattice-orgs` has it.

src/containers/studies/components/CreateStudyModal.js
```diff
@@ -0,0 +1,111 @@
...
+        <ModalBodyWrapper>
+          <CreateStudyForm
+              formData={data}
+              handleOnChange={this.handleOnChange}
+              isSubmitting={isSubmitting} />
+        </ModalBodyWrapper>
+      ),
+      [RequestStates.FAILURE]: (
+        <ModalBodyWrapper>
+          <span> Failed to create a new study. Please try again </span>
```
UnsungHero97: You need a period at the end of `Please try again.`.

UnsungHero97 (changes requested):

src/containers/studies/components/CreateStudyForm.js
solomontang: You should also the data processing utils to handle your data. Unfortunately the documentation isn't very good but there is a working example of how to set up the schema and such on storybook

https://openlattice.github.io/lattice-fabricate/?path=/story/form--data-processing-w-edits-delete

src/containers/studies/components/CreateStudyModal.js
```diff
@@ -0,0 +1,108 @@
...
+    this.setState({
+      data: {
+        description: formData.description || '',
+        email: formData.email || '',
+        group: formData.group || '',
+        name: formData.name || '',
+        version: formData.version || ''
+      }
+    });
+  }
```
solomontang: You only need to supply an `onChange` if you want to execute side effects (this that occur outside of updating form field states). Since this callback function is only controlling the state which is already managed by the Form component internally, all this will do is incur extraneous render cycles.

You can track this by dropping a `console.log` in the `render` method of `CreateStudyForm` and you should see that it is triggered on every field update. By omitting this function, you should still have the same behavior without re-rendering the entire parent component.

src/containers/studies/components/CreateStudyForm.js
```diff
@@ -0,0 +1,70 @@
...
+}
+
+const CreateStudyForm = ({ formData, handleOnChange, isSubmitting } :Props) => (
+  <Form
+      formData={formData}
+      hideSubmit
+      isSubmitting={isSubmitting}
+      onChange={handleOnChange}
+      schema={dataSchema}
+      uiSchema={uiSchema} />
```
solomontang: Fabricate is built on top of the `react-jsonschema-form` library and for the basics, you can use their website for documention. Our internal widgets try to match what is provided out of the box from this library. However, you're probably going to need a rundown of how the `DataProcessingUtils` work.

You do not need to provide an `onChange` prop to the `Form` if you are not executing any side effects.
https://react-jsonschema-form.readthedocs.io/en/latest/#form-initialization

Please provide an onSubmit
https://react-jsonschema-form.readthedocs.io/en/latest/#form-initialization

I suspect CreateStudyForm will not need to prepopulate any data and thus the `formData` prop can also be omitted.
https://react-jsonschema-form.readthedocs.io/en/latest/#form-initialization

anzioka:
modal

Added functionality to create a new study by leveraging fabricate's DataProcessingUtils. The shiny new study form includes input validation (name and email are required fields).

anzioka:
Screen Shot 2019-12-11 at 5 44 08 PM

src/containers/studies/StudiesContainer.js
```diff
@@ -55,24 +57,50 @@ const CenterText = styled.div`
...
-    // TODO: open modal to create study
+    this.setState({
+      isCreateStudyModalVisible: true
+    });
+  }
+  handleOnCloseModal = () => {
+    const { actions } = this.props;
+
+    // this ensures that the modal always renders CreateStudyForm when opened
+    actions.resetRequestState(CREATE_STUDY);
```
solomontang: Judging from when this action is dispatched, I suspect you might have a visual bug.

As the form is fading out, does the modal view also reset back to its standby state with an empty form instead of the success message? This is probably happening since the state is reset before the animation can finish. Ideally, you would couple the `resetRequestState` when the contents of the Modal are unmounted.

src/containers/studies/components/CreateStudyForm.js
```diff
@@ -0,0 +1,58 @@
...
+      [getPageSectionKey(1, 1), getEntityAddressKey(0, CHRONICLE_STUDIES, STUDY_ID)], uuid());
+
+    const entityData = processEntityData(newFormData, allEntitySetIds, propertyTypeIds);
+    const assocationEntityData = {};
+
+    dispatch(createStudy({
+      entityData,
+      assocationEntityData
+    }));
+  });
```
solomontang: Yay hooks!

This call to `useCallback` does not have a dependency list. This means that the callback is not really memoized since it will change upon every render. I recommend removing `useCallback` until you actually observe performance consequences or bad render cycles.

I suspect you should have an eslint error if you have installed the necessary plugins. You should do your best to heed the linter when doing anything Hook related as it will stop an entire class of bugs ahead of time.

References:
https://reactjs.org/docs/hooks-reference.html#usecallback
https://www.npmjs.com/package/eslint-plugin-react-hooks

src/containers/studies/components/CreateStudyModal.js
```diff
@@ -0,0 +1,94 @@
...
+const CreateStudyModal = (props :Props) => {
+  const formRef = useRef();
+
+  const { isVisible, handleOnCloseModal, requestStates } = props;
+
+  const handleOnSubmit = useCallback(() => {
+    if (formRef.current) {
+      formRef.current.submit();
+    }
+  });
```
solomontang: Same as the previous comment about `useCallback`. Either provide the exhaustive dependency array, or don't use `useCallback`.

solomontang (changes requested):

src/containers/studies/components/CreateStudyForm.js
solomontang: As a follow-up clarification, lifecycle methods are concerned with when some change should occur. `componentDidMount`, `componentDidUpdate`, `componentWillUnmount` are event driven. Often times, this leads to repeated code across different lifecycle methods and manual checks for equality/change.

Hooks on the other hand are concerned with state synchronization. When you provide a dependency list, you're declaring that your effect/memo/callback is in sync with those specific items by reference equality. By not providing a dependency list, you're saying that your effect/memo/callback is in sync with everything and will update on every render. Similarly, providing an empty dependency list means your hook synchronizes with nothing and will only be run once.

In short, using hooks can be beneficial and fun, but it does take changing the way you have previously approached React.

src/containers/studies/components/CreateStudyModal.js
```diff
@@ -0,0 +1,94 @@
...
+        </ModalBodyWrapper>
+      ),
+      [RequestStates.SUCCESS]: (
+        <ModalBodyWrapper>
+          <span> Successfully created a new study. </span>
+        </ModalBodyWrapper>
+      )
+    };
+    return components;
+  };
```
UnsungHero97: This doesn't need to be a function, `requestStateComponents` can be a plain object.

src/containers/studies/components/CreateStudyModal.js
```diff
@@ -0,0 +1,94 @@
...
+type Props = {
+  handleOnCloseModal :() => void;
+  isVisible :boolean;
+  actions:{
+    createStudy :RequestSequence
+  };
+  requestStates:{
+    CREATE_STUDY :RequestState
+  };
+};
```
UnsungHero97: Please sort the props alphabetically.

src/containers/studies/components/CreateStudyModal.js
```diff
@@ -0,0 +1,94 @@
+/*
+ * @flow
+ */
+import React, { useCallback, useRef } from 'react';
+
+import styled from 'styled-components';
+import { Map } from 'immutable';
+import {
+  ActionModal
+} from 'lattice-ui-kit';
```
UnsungHero97: You only need to split into multiple lines if there's 4 or more imports.

src/core/edm/EDMReducer.js
```diff
@@ -112,13 +117,15 @@ export default function edmReducer(state :Map<*, *> = INITIAL_STATE, action :Obj
             .set('entityTypes', entityTypes.asImmutable())
             .set('entityTypesIndexMap', entityTypesIndexMap.asImmutable())
             .set('propertyTypes', propertyTypes.asImmutable())
+            .set('propertyTypesFqnIdMap', propertyTypesFqnIdMap.asImmutable())
```
UnsungHero97, line 122: You haven't defined this in `INITIAL_STATE`.

src/core/edm/EDMReducer.js
```diff
@@ -98,6 +100,9 @@ export default function edmReducer(state :Map<*, *> = INITIAL_STATE, action :Obj
                 .setTitle(pt.title)
                 .setType(pt.type)
                 .build();
+
+              const typeFqn :FullyQualifiedName = new FullyQualifiedName(propertyType.type);
```
UnsungHero97: Use the shorter `FQN` annotation:

```
import type { FQN } from 'lattice';
const fqn :FQN = new FullyQualifiedName(propertyType.type);
```

src/containers/studies/components/CreateStudyForm.js
```diff
@@ -0,0 +1,58 @@
...
+  const propertyTypeIds :Map = useSelector((store :Map) => store.getIn(['edm', 'propertyTypesFqnIdMap']));
+  const dispatch = useDispatch();
+
+  const handleSubmit = useCallback((payload :any) => {
+    let { formData: newFormData } = payload;
+    newFormData = setIn(newFormData,
+      [getPageSectionKey(1, 1), getEntityAddressKey(0, CHRONICLE_STUDIES, STUDY_ID)], uuid());
+
+    const entityData = processEntityData(newFormData, allEntitySetIds, propertyTypeIds);
+    const assocationEntityData = {};
```
UnsungHero97: This is misspelled; should be `associationEntityData`.

src/core/sagas/data/DataSagas.js
```diff
@@ -0,0 +1,71 @@
...
+  if (value === null || value === undefined) {
+    workerResponse.error = ERR_ACTION_VALUE_NOT_DEFINED;
+    yield put(submitDataGraph.failure(id, workerResponse.error));
+    return workerResponse;
+  }
+  try {
+    yield put(submitDataGraph.request(id, value));
+
+    const dataGraph = (new DataGraphBuilder())
+      .setAssociations(value.assocationEntityData)
```
UnsungHero97: This is misspelled; should be `associationEntityData`.

src/containers/studies/components/CreateStudyForm.js
```diff
@@ -0,0 +1,58 @@
...
+
+    dispatch(createStudy({
+      entityData,
+      assocationEntityData
```
UnsungHero97: This is misspelled; should be `associationEntityData`.

src/containers/studies/components/CreateStudyForm.js
```diff
@@ -0,0 +1,58 @@
...
+const CreateStudyForm = (props :Props, ref) => {
+
+  const allEntitySetIds :Map = useSelector((store :Map) => store.getIn(['edm', 'entitySetIds']));
+  const propertyTypeIds :Map = useSelector((store :Map) => store.getIn(['edm', 'propertyTypesFqnIdMap']));
+  const dispatch = useDispatch();
+
+  const handleSubmit = useCallback((payload :any) => {
+    let { formData: newFormData } = payload;
+    newFormData = setIn(newFormData,
+      [getPageSectionKey(1, 1), getEntityAddressKey(0, CHRONICLE_STUDIES, STUDY_ID)], uuid());
```
UnsungHero97: Please fix formatting:

```
newFormData = setIn(
  newFormData,
  [getPageSectionKey(1, 1), getEntityAddressKey(0, CHRONICLE_STUDIES, STUDY_ID)],
  uuid(),
);
```

src/containers/studies/components/CreateStudyForm.js
```diff
@@ -0,0 +1,58 @@
...
+import { ENTITY_SET_NAMES } from '../../../core/edm/constants/EntitySetNames';
+import { PROPERTY_TYPE_FQNS } from '../../../core/edm/constants/FullyQualifiedNames';
+import { createStudy } from '../StudiesActions';
+
+const { STUDY_ID } = PROPERTY_TYPE_FQNS;
+const { CHRONICLE_STUDIES } = ENTITY_SET_NAMES;
+const { processEntityData, getEntityAddressKey, getPageSectionKey } = DataProcessingUtils;
+
+type Props = {
+};
```
UnsungHero97: Why is this empty?

src/containers/studies/StudiesContainer.js
```diff
@@ -109,21 +137,26 @@ class StudiesContainer extends Component<Props> {
...
       </>
     );
   }
 }
 const mapStateToProps = (state :Map) => ({
   requestStates: {
-    [GET_STUDIES]: state.getIn(['studies', GET_STUDIES, 'requestState'])
+    [GET_STUDIES]: state.getIn(['studies', GET_STUDIES, 'requestState']),
+    [CREATE_STUDY]: state.getIn(['studies', CREATE_STUDY, 'requestState']),
+
```
UnsungHero97: Please remove empty line.

src/containers/studies/StudiesReducer.js
```diff
@@ -31,6 +42,14 @@ export default function studiesReducer(state :Map<*, *> = INITIAL_STATE, action
       });
     }
 
+    case createStudy.case(action.type): {
+      return createStudy.reducer(state, action, {
+        REQUEST: () => state.setIn([CREATE_STUDY, 'requestState'], RequestStates.PENDING),
+        SUCCESS: () => state.setIn([CREATE_STUDY, 'requestState'], RequestStates.SUCCESS),
```
UnsungHero97: On success, you need to update state with the new study so that it shows up in the list.

src/containers/studies/StudiesSagas.js
```diff
@@ -50,7 +56,36 @@ function* getStudiesWatcher() :Generator<*, *, *> {
   yield takeEvery(GET_STUDIES, getStudiesWorker);
 }
 
+
+function* createStudyWorker(action :SequenceAction) :Generator<*, *, *> {
+  const { id, value } = action;
+  let response = {};
```
UnsungHero97: This doesn't need to be a `let`. It can be a `const` on line 66.

src/containers/studies/StudiesContainer.js
anzioka: You are right! There was a visual bug, though almost unnoticeable. The success message was rendered but on closing the modal an empty form would be temporarily rendered before the modal closed. I fixed the issue by calling resetRequestState before the modal was rendered (when the isVisible prop is set to true).

src/containers/studies/StudiesSagas.js
```diff
@@ -50,7 +56,36 @@ function* getStudiesWatcher() :Generator<*, *, *> {
...
+
+    response = yield call(submitDataGraphWorker, submitDataGraph(value));
+    if (response.error) {
+      throw response.error;
+    }
+
+    yield put(createStudy.success(id));
+  }
+  catch (error) {
+    LOG.error('caught exception in createStudy()', error);
```
UnsungHero97: This should be `LOG.error(action.type)`. Please follow the patterns consistently.

src/containers/studies/components/CreateStudyModal.js
```diff
@@ -0,0 +1,94 @@
...
+import type { RequestSequence, RequestState } from 'redux-reqseq';
+
+import CreateStudyForm from './CreateStudyForm';
+
+import { CREATE_STUDY, createStudy } from '../StudiesActions';
+
+type Props = {
+  handleOnCloseModal :() => void;
+  isVisible :boolean;
+  actions:{
```
UnsungHero97: There should be a space before the colon... `actions :{`

src/containers/studies/components/CreateStudyModal.js
```diff
@@ -0,0 +1,94 @@
...
+
+const mapStateToProps = (state :Map) => ({
+  requestStates: {
+    [CREATE_STUDY]: state.getIn(['studies', CREATE_STUDY, 'requestState'])
+  },
+});
+
+const mapDispatchToProps = (dispatch :Function) => ({
+  actions: bindActionCreators({
+    createStudy,
```
UnsungHero97: This action is not used in the component.

UnsungHero97 (changes requested):

src/containers/studies/components/CreateStudyModal.js
```diff
@@ -16,16 +14,15 @@ import type { RequestSequence, RequestState } from 'redux-reqseq';
...
-import { CREATE_STUDY, createStudy } from '../StudiesActions';
+import { CREATE_STUDY } from '../StudiesActions';
 
 type Props = {
-  handleOnCloseModal :() => void;
-  isVisible :boolean;
-  actions:{
-    createStudy :RequestSequence,
+  actions :{
     resetRequestState :RequestSequence
```
UnsungHero97, line 19: `resetRequestState` is also not used anywhere. Please remove.

src/containers/studies/components/CreateStudyModal.js
```diff
@@ -93,7 +87,6 @@ const mapStateToProps = (state :Map) => ({
 
 const mapDispatchToProps = (dispatch :Function) => ({
   actions: bindActionCreators({
-    createStudy,
     resetRequestState
   }, dispatch)
 });
```
UnsungHero97: Given that the component doesn't use any actions, it's safe to remove `mapDispatchToProps` entirely.

UnsungHero97 (changes requested):

package.json
```diff
@@ -77,6 +79,7 @@
     "eslint-plugin-jest": "~23.1.0",
     "eslint-plugin-jsx-a11y": "~6.2.0",
     "eslint-plugin-react": "~7.17.0",
+    "eslint-plugin-react-hooks": "^2.3.0",
```
UnsungHero97: Please use `~` instead of `^`.

solomontang: Pin the version to patch by using `~` instead of `^`

src/containers/studies/StudiesContainer.js
```diff
@@ -55,24 +57,51 @@ const CenterText = styled.div`
...
   componentDidMount() {
     const { actions } = this.props;
     actions.getStudies();
   }
+
   openCreateStudyModal = () => {
-    // TODO: open modal to create study
+    const { actions } = this.props;
+    // necessary after a successful or failed CREATE_STUDY action
+    actions.resetRequestState(CREATE_STUDY);
```
UnsungHero97, line 89: Why do you trigger this action before the modal opens?

src/containers/studies/StudiesContainer.js
```diff
@@ -109,21 +138,26 @@ class StudiesContainer extends Component<Props> {
...
+
   },
   studies: state.getIn(['studies', 'studies'], List())
 });
 
 const mapDispatchToProps = (dispatch :Function) => ({
   actions: bindActionCreators({
     getStudies,
+    resetRequestState,
     goToRoute: RoutingActions.goToRoute,
```
solomontang: You should get in the habit of sorting. We're probably going to be using prettier soon so this will be less of an ordeal in the future.

src/containers/studies/StudiesReducer.js
```diff
@@ -26,11 +37,26 @@ export default function studiesReducer(state :Map<*, *> = INITIAL_STATE, action
...
       });
     }
 
+    case createStudy.case(action.type): {
+      return createStudy.reducer(state, action, {
+        REQUEST: () => state.setIn([CREATE_STUDY, 'requestState'], RequestStates.PENDING),
+        FAILURE: () => state.setIn([CREATE_STUDY, 'requestState'], RequestStates.FAILURE),
+        SUCCESS: () => {
+          const { studyUUID } = action.value;
+          const study = state.getIn(['studies', studyUUID], Map());
```
UnsungHero97: I'd be surprised if this is actually working. In the worker, you have:

```
yield put(createStudy.success(id));
```

This means `action.value` is equal to `{}` and `studyUUID` will be undefined.

Separately, why are you getting the new study from state? It's a new study, which means it won't be in state, which means you'll always be storing an empty map below.

src/containers/studies/StudiesSagas.js
```diff
@@ -50,7 +56,35 @@ function* getStudiesWatcher() :Generator<*, *, *> {
   yield takeEvery(GET_STUDIES, getStudiesWorker);
 }
 
+
+function* createStudyWorker(action :SequenceAction) :Generator<*, *, *> {
+  const { id, value } = action;
```
UnsungHero97: Please move this inside the `try` block.

UnsungHero97 (changes requested):

src/containers/studies/StudiesContainer.js
solomontang: I suspect this is to clear any previous lingering request states. Before, he was clearing on close but since it was being called in the parent container, it was causing flickering when then modal was transitioning out. 

Ideally, you would co-locate as much of the logic relating to the contents of your modal all in one place. The Modal will unmount its children when the transition is complete. So as long as the contents of the modal are responsible for their own cleanup, there won't be any issues.

But this seems to be the next best way.

src/containers/studies/StudiesContainer.js
anzioka: Ok, sorted.

src/containers/studies/StudiesReducer.js
anzioka: I fixed this.

src/containers/studies/StudiesContainer.js
anzioka: You are right @solomontang. After a SUCCESS/ FAILURE success state, opening the modal a second time would not render the form component, that's why I reset the state before the modal opens.

anzioka:
Merged  feature/APPS-1592-create-study-details-page with this branch.

PR #6, opened 2019-12-18
Feature/apps 1592 create study details page

anzioka:
participant
study_details

Add participant to study (top image)
Display study details (study details - name, version, description - will ultimately be editable. Functionality not included for now.


src/core/edm/EDMSagas.js
```diff
@@ -111,10 +124,90 @@ function* getAllEntitySetIdsWatcher() :Generator<*, *, *> {
...
 }
 
+function* getParticipantsEntitySetsIdsWorker(action :SequenceAction) :Generator<*, *, *> {
+  const workerResponse = {};
+  try {
+    yield put(getParticipantsEntitySetsIds.request(action.id));
+    const studies = yield select((state) => state.getIn(['studies', 'studies'], Map()));
+    const studyIds = studies
+      .valueSeq()
+      .map((entry :Map) => entry.getIn([STUDY_ID, 0]))
```
UnsungHero97: `entry` is misleading. Please rename to either `study` or `value`.

src/core/edm/EDMSagas.js
```diff
@@ -111,10 +124,90 @@ function* getAllEntitySetIdsWatcher() :Generator<*, *, *> {
...
+      .map((studyId :string) => `${PARTICIPANTS_PREFIX}${studyId}`);
+
+    let responseData = {};
+
+    if (!studyIds.isEmpty()) {
+      const response = yield call(getEntitySetIdsWorker, getEntitySetIds(studyIds.toJS()));
+      if (response.error) throw response.error;
+      responseData = response.data;
+    }
```
UnsungHero97: Unfortunately, this won't work. `getEntitySetIds` will fail if you're requesting the id of an entity set you don't have access to, which will be very common.

As a result, if `getEntitySetIds` fails, then `getParticipantsEntitySetsIdsWorker` will fail, which means `initializeApplicationWorker` will fail, which means the app will not initialize.

You should get participant entity sets only when you're going to view a specific study's details.

src/core/router/RouterUtils.js
```diff
@@ -0,0 +1,9 @@
+/*
+ * @flow
+ */
+
+import type { Match } from 'react-router';
+
+const getIdFromMatch = (match :Match) => match.params.id;
```
UnsungHero97: This needs to be more bulletproof. Here's how I've done this...

```
const getIdFromMatch = (match :Match) :?string => {

  const {
    params: {
      id = null,
    } = {},
  } = match;

  return id;
};
```

src/core/edm/EDMSagas.js
```diff
@@ -111,10 +124,90 @@ function* getAllEntitySetIdsWatcher() :Generator<*, *, *> {
...
 export {
+  createParticipantsEntitySetWatcher,
+  createParticipantsEntitySetWorker,
+  getAllEntitySetIdsWatcher,
+  getAllEntitySetIdsWorker,
   getEntityDataModelTypesWatcher,
   getEntityDataModelTypesWorker,
-  getAllEntitySetIdsWorker,
-  getAllEntitySetIdsWatcher
+  getParticipantsEntitySetsIdsWorker
```
UnsungHero97: These don't belong in `EDMSagas`:

```
createParticipantsEntitySetWatcher
createParticipantsEntitySetWorker
getParticipantsEntitySetsIdsWorker
```

They belong in `StudiesSagas` (or `StudySagas` if you decide to create that).

src/core/edm/EDMSagas.js
```diff
@@ -111,10 +124,90 @@ function* getAllEntitySetIdsWatcher() :Generator<*, *, *> {
...
+
+    const entityTypeId = yield call(getEntityTypeId, PERSON);
+    const entitySet = {
+      entityTypeId,
+      name: `${PARTICIPANTS_PREFIX}${studyId}`,
+      title: `${studyName} Participants`,
+      description: `Participants of study with name ${studyName} and id ${studyId}`,
+      external: false,
+      contacts: [email]
+    };
```
UnsungHero97: Please use `EntitySetBuilder` from `lattice`. Take a look at the `EDMReducer` as a reference:

https://github.com/openlattice/lattice-orgs/blob/develop/src/core/edm/EDMReducer.js#L143

src/containers/app/AppSagas.js
```diff
@@ -38,7 +47,16 @@ function* initializeApplicationWorker(action :SequenceAction) :Generator<*, *, *
...
-    if (responses[0].error) throw responses[0].error;
+    responses.forEach((res) => {
+      if (res.error) throw res.error;
+    });
+    // get all studies only after getting entitySetIds
+    let response = yield call(getStudiesWorker, getStudies());
+    if (response.error) throw response.error;
+
+    response = yield call(getParticipantsEntitySetsIdsWorker, getParticipantsEntitySetsIds());
+    if (response.error) throw response.error;
```
UnsungHero97: It's a good idea to take this out of the initialization step and only load participant entity sets when looking at a specific study.

src/containers/studies/StudiesActions.js
```diff
@@ -5,10 +5,20 @@
...
 import type { RequestSequence } from 'redux-reqseq';
 
+const ADD_PARTICIPANT :'ADD_PARTICIPANT' = 'ADD_PARTICIPANT';
+const addStudyParticipant :RequestSequence = newRequestSequence(ADD_PARTICIPANT);
+
 const GET_STUDIES :'GET_STUDIES' = 'GET_STUDIES';
 const getStudies :RequestSequence = newRequestSequence(GET_STUDIES);
 
+const CREATE_STUDY :'CREATE_STUDY' = 'CREATE_STUDY';
+const createStudy :RequestSequence = newRequestSequence(CREATE_STUDY);
```
UnsungHero97, line 15: Please sort these alphabetically:

```
const ADD_PARTICIPANT ...
const addStudyParticipant

const CREATE_STUDY ...
const createStudy ...

const GET_STUDIES ...
const getStudies ...
```

src/containers/studies/StudiesSagas.js
```diff
@@ -1,56 +1,149 @@
...
 
+    const response = yield call(getEntitySetDataWorker, getEntitySetData({ entitySetId }));
     if (response.error) {
       throw response.error;
     }
 
-    yield put(getStudies.success(action.id, response.data));
+    const studies = {};
+    response.data.forEach((study) => {
+      studies[study[OPENLATTICE_ID_FQN]] = study;
```
UnsungHero97: Two things:

1. don't forget `study[OPENLATTICE_ID_FQN]` is an array. You need to grab the 0th index:
```
studies[study[OPENLATTICE_ID_FQN, 0]] = study;
```

2. I feel like this is a bit cleaner / easier:
```
const studies :Map<UUID, Map> = fromJS(response.data)
  .toMap()
  .mapKeys((index :number, study :Map) => study.get('id'));
```

UnsungHero97 (changes requested):

src/core/edm/EDMReducer.js
```diff
@@ -98,6 +102,9 @@ export default function edmReducer(state :Map<*, *> = INITIAL_STATE, action :Obj
                 .setTitle(pt.title)
                 .setType(pt.type)
                 .build();
+
+              const typeFqn :FQN = new FullyQualifiedName(propertyType.type);
+              propertyTypesFqnIdMap.set(typeFqn, propertyType.id);
```
UnsungHero97: You don't need to do `new FullyQualifiedName` here because `propertyType.type` is already an FQN.

src/containers/studies/StudiesReducer.js
```diff
@@ -26,10 +58,59 @@ export default function studiesReducer(state :Map<*, *> = INITIAL_STATE, action
...
+    }
+
+    case addStudyParticipant.case(action.type): {
+      const seqAction :SequenceAction = action;
+      return addStudyParticipant.reducer(state, action, {
+        REQUEST: () => state.setIn([ADD_PARTICIPANT, 'requestState'], RequestStates.PENDING),
+        FAILURE: () => state.setIn([ADD_PARTICIPANT, 'requestState'], RequestStates.FAILURE),
+        SUCCESS: () => {
+          const { participant } = seqAction.value;
+          const { studyId } = seqAction.value;
```
UnsungHero97: First, you can destructure more than one field at a time:

```
const { participant, studyId } = seqAction.value;
```

Second, based on what I see in the worker, this won't work.

```
yield put(addStudyParticipant.success(id, response.data));
```

When you do `addStudyParticipant.success(id, response.data)`, you are setting `seqAction.value` to equal `response.data`. As such, `seqAction.value` will not have `participant` or `studyId`.

src/containers/studies/StudiesReducer.js
```diff
@@ -26,10 +58,59 @@ export default function studiesReducer(state :Map<*, *> = INITIAL_STATE, action
...
+      const seqAction :SequenceAction = action;
+      return createParticipantsEntitySet.reducer(state, action, {
+        REQUEST: () => state.setIn([CREATE_PARTICIPANTS_ENTITY_SET, 'requestState'], RequestStates.PENDING),
+        FAILURE: () => state.setIn([CREATE_PARTICIPANTS_ENTITY_SET, 'requestState'], RequestStates.FAILURE),
+        SUCCESS: () => {
+          const { entitySetName, entitySetId } = seqAction.value;
+          const updatedMap = state.get('participantEntitySetIds').set(entitySetName, entitySetId);
+          return state
+            .setIn([CREATE_PARTICIPANTS_ENTITY_SET, 'requestState'], RequestStates.SUCCESS)
+            .set('participantEntitySetIds', updatedMap);
```
UnsungHero97, line 155: You have not defined `participantEntitySetIds` in `INITIAL_STATE`.

src/containers/studies/StudiesSagas.js
```diff
@@ -1,56 +1,207 @@
...
+    yield put(createParticipantsEntitySet.request(action.id));
+    const newStudyData = action.value;
+    const studyName = getIn(newStudyData,
+      [getPageSectionKey(1, 1), getEntityAddressKey(0, CHRONICLE_STUDIES, STUDY_NAME)]);
+    const studyId = getIn(newStudyData,
+      [getPageSectionKey(1, 1), getEntityAddressKey(0, CHRONICLE_STUDIES, STUDY_ID)]);
+    const email = getIn(newStudyData,
+      [getPageSectionKey(1, 1), getEntityAddressKey(0, CHRONICLE_STUDIES, STUDY_EMAIL)]);
+
+    const entityTypeId = yield call(getEntityTypeId, PERSON);
```
UnsungHero97: You don't need to make an API call for this. The data is already in redux.

```
const entityTypeId :UUID = yield select((state :Map) => {
  const entityTypeIndex :number = state.getIn(['edm', 'entityTypesIndexMap', PERSON]);
  return state.getIn(['edm', 'entityTypes', entityTypeIndex, 'id']);
});
```

In fact, you can simplify this further with a simple helper...

```
// src/core/edm/EDMUtils.js
/*
 * @flow
 */

import { Map } from 'immutable';
import type { FQN } from 'lattice';

const selectEntityTypeId = (entityTypeFQN :FQN) => {
  return (state :Map) => {
    const entityTypeIndex :number = state.getIn(['edm', 'entityTypesIndexMap', entityTypeFQN]);
    return state.getIn(['edm', 'entityTypes', entityTypeIndex, 'id']);
  };
};

export {
  selectEntityTypeId,
};



// src/containers/studies/StudiesSagas.js
...
const entityTypeId :UUID = yield select(selectEntityTypeId(PERSON));
...
```

src/containers/studies/StudiesSagas.js
```diff
@@ -1,56 +1,207 @@
...
+    }
+
+    // get the created entity
+    const { entityKeyIds } = response.data;
+    const entitySetId = yield select(
+      (state) => state.getIn(['edm', 'entitySetIds', CHRONICLE_STUDIES])
+    );
+    const entityKeyId = entityKeyIds[entitySetId][0];
+    response = yield call(getEntityDataWorker, getEntityData({ entityKeyId, entitySetId }));
+    if (response.error) throw response.error;
```
UnsungHero97: Why do you need to make this API call? You already have the study details in `action.value`, and the response from `submitDataGraph` gives you the entity key id, so you should have everything you need.

UnsungHero97 (changes requested):

src/containers/studies/StudiesSagas.js
```diff
@@ -1,56 +1,225 @@
...
+  // do not create a new study if createParticipantsEntitySet fails
+  try {
+    const { id, value } = action;
+    let { newStudyData } = value;
+    let response = {};
+
+    response = yield call(createParticipantsEntitySetWorker, createParticipantsEntitySet(newStudyData));
+    if (response.error) throw response.error;
+
+    yield put(createStudy.request(id, value));
```
UnsungHero97: `createStudy.request` MUST be the first line of code inside the `try` block:

```
function* createStudyWorker(action :SequenceAction) :Generator<*, *, *> {
  try {
    yield put(createStudy.request(action.id, action.value));
  ...
}
```

src/containers/studies/StudiesReducer.js
```diff
@@ -26,10 +78,83 @@ export default function studiesReducer(state :Map<*, *> = INITIAL_STATE, action
...
+            [STUDY_EMAIL.toString()]:
+              [get(pageSection, getEntityAddressKey(0, CHRONICLE_STUDIES, STUDY_EMAIL))],
+            [OPENLATTICE_ID_FQN.toString()]:
+              [get(pageSection, getEntityAddressKey(0, CHRONICLE_STUDIES, OPENLATTICE_ID_FQN))],
+          };
+          const updatedStudies = state.get('studies').set(study[STUDY_ID.toString()], fromJS(study));
+
+          return state
+            .set('studies', updatedStudies)
+            .setIn([CREATE_STUDY, 'requestState'], RequestStates.SUCCESS);
```
UnsungHero97: A quick note, you don't need to do `.toString()` on the FQNs. It *should* just work.

Now, what if you had to do this 1000 times? What if you had to do this 1000 times and the entity data is different each time? Would you take the same approach?

UnsungHero97 (changes requested):

UnsungHero97 (approved):

PR #13, opened 2020-01-08
Feature/apps 1593 study participants table

anzioka:
Add study participants
add_participant

Change participant enrollment status
change_enrollment

Delete participant
delete_particant

Show participant 
participant_info
info

participants table
participants_table


src/core/edm/constants/DataModelConstants.js
```diff
@@ -0,0 +1,26 @@
+// @flow
+
+// study enrollment status
+const ENROLLMENT_STATUS = {
+  ENROLLED: 'enrolled',
+  NOT_ENROLLED: 'not_enrolled'
+};
```
UnsungHero97: Two things:

1. Please follow this enum pattern: 
https://github.com/openlattice/lattice-js/blob/develop/src/constants/types/ActionTypes.js

2. The server expects the status in all-caps, so `enrolled` and `not_enrolled` won't work. See here:
https://github.com/openlattice/chronicle-server/blob/develop/src/main/java/com/openlattice/chronicle/constants/ParticipationStatus.java

src/core/edm/constants/DataModelConstants.js
```diff
@@ -0,0 +1,26 @@
+// @flow
```
UnsungHero97: The constants in this file are not data model constants. As such, I think the name of this file is misleading and it doesn't belong in `core/edm`.

src/utils/constants/UrlConstants.js
```diff
@@ -0,0 +1,21 @@
+// @flow
+
+const CHRONICLE :string = 'chronicle';
+const DATA :string = 'data';
+const FILE_TYPE :string = 'fileType';
+const LOCAL :string = 'local';
+const PARTICIPANT :string = 'participant';
+const PRODUCTION :string = 'production';
+const STAGING :string = 'staging';
```
UnsungHero97: `LOCAL`, `PRODUCTION`, and `STAGING` are environment types, and I think they should follow the enum pattern, like `ENROLLMENT_STATUS`.

src/utils/api/AppApi.js
```diff
@@ -0,0 +1,49 @@
...
+} from '../constants/UrlConstants';
+
+declare var __ENV_DEV__ :boolean; // injected by Webpack.DefinePlugin
+
+const LOG = new Logger('AppApi');
+
+const getBaseUrl = () => {
+  if (__ENV_DEV__) {
+    return ENV_URLS.get(LOCAL);
+  }
```
UnsungHero97: Let's ignore webpack and get rid of `__ENV_DEV__` for this function. You can use this as a reference:

https://github.com/openlattice/lattice-auth/blob/develop/src/config/Configuration.js#L40

UnsungHero97 (changes requested):

src/core/edm/constants/DataModelConstants.js
anzioka: Moved to utils/constants/GlobalConstants.js

src/containers/studies/StudiesSagas.js
```diff
@@ -10,63 +10,332 @@ import {
...
 
 const LOG = new Logger('StudiesSagas');
 
+/*
+ *
+ * StudiesActions.changeEnrollmentStatus()
+ *
+ */
+
+function* changeEnrollmentWorker(action :SequenceAction) :Generator<*, *, *> {
```
UnsungHero97: Rename to `changeEnrollmentStatusWorker`.

src/containers/studies/StudiesSagas.js
```diff
@@ -10,63 +10,332 @@ import {
...
+  catch (error) {
+    LOG.error(action.type, error);
+    yield put(changeEnrollmentStatus.failure(action.id));
+  }
+  finally {
+    yield put(changeEnrollmentStatus.finally(action.id));
+  }
+}
+
+function* changeEnrollmentWatcher() :Generator<*, *, *> {
```
UnsungHero97: Rename to `changeEnrollmentStatusWatcher`.

src/containers/studies/StudiesSagas.js
```diff
@@ -10,63 +10,332 @@ import {
...
+
+    const participantsEntityName = `${PARTICIPANTS_PREFIX}${studyId}`;
+    const participantsEntitySetId = yield select(
+      (state) => state.getIn(['studies', 'participantEntitySetIds', participantsEntityName])
+    );
+    const associationEntityKeyId = yield select(
+      (state) => state.getIn(['studies', 'associationKeyIds', participantsEntitySetId, participantEntityKeyId])
+    );
+    const statusPropertyTypeId = yield select((state) => state.getIn(['edm', 'propertyTypeIds', STATUS]));
+    const participatedInEntitySetId = yield select((state) => state.getIn(['edm', 'entitySetIds', PARTICIPATED_IN]));
```
UnsungHero97: This is totes fine. Just sharing a thought.

I prefer the following because it minimizes the `yield select`:

```
const { thing1, thing2 } = yield select((state) => ({
  thing1: state.getIn(...),
  thing2: state.getIn(...),
  ...
}));
```

This can also be refactored into a helper that returns many things from state, and then the destructuring will pick out what you care about.

src/containers/studies/StudiesSagas.js
```diff
@@ -10,63 +10,332 @@ import {
...
+    yield put(deleteStudyParticipant.failure(action.id));
+  }
+  finally {
+    yield put(deleteStudyParticipant.finally(action.id));
+  }
+}
+
+function* deleteStudyParticipantWatcher() :Generator<*, *, *> {
+  yield takeEvery(DELETE_STUDY_PARTICIPANT, deleteStudyParticipantWorker);
+}
```
UnsungHero97: It is not enough to just delete the participant from the entity set. You also need to delete all the relevant data that is collected for the participant. You need to rewrite this entirely and model it after `hardDeleteParticipantData` in the old app.

src/containers/studies/StudiesSagas.js
```diff
@@ -10,63 +10,332 @@ import {
...
+      const enrollmentStatus :Map = fromJS(response.data)
+        .map((associations :List) => associations.first().getIn(['associationDetails', STATUS, 0]));
+      workerResponse.data = enrollmentStatus;
+
+      // mapping from participantEntityKeyId -> association EKID
+      const associationKeyIds :Map = fromJS(response.data)
+        .map((associations :List) => associations.first().getIn(['associationDetails', OPENLATTICE_ID_FQN, 0]));
+
+      yield put(getParticipantsEnrollmentStatus.success(action.id, { associationKeyIds, participantsEntitySetId }));
+    }
```
UnsungHero97: You are calling `getParticipantsEnrollmentStatus.success()` only when `if (!participants.isEmpty())` is true. In the case that it's false, `.success()` is never invoked, which means the success reducer will not run.

src/containers/studies/StudiesSagas.js
```diff
@@ -10,63 +10,332 @@ import {
...
+
+function* getStudyParticipantsWorker(action :SequenceAction) :Generator<*, *, *> {
+  try {
+    yield put(getStudyParticipants.request(action.id));
+
+    const studyId = action.value;
+
+    // only fetch if data is not found in redux store.
+    let participants = yield select((state) => state.getIn(['studies', 'participants', studyId]));
+    if (participants) {
```
UnsungHero97: What are your reasons for conditionally fetching participants? What use case requires this?

src/containers/studies/StudiesSagas.js
```diff
@@ -199,20 +468,54 @@ function* addStudyParticipantWorker(action :SequenceAction) :Generator<*, *, *>
...
+    const entitySetName = `${PARTICIPANTS_PREFIX}${studyId}`;
+    const participantsEntitySetId = participantEntitySetIds.get(entitySetName);
+
+    const associations = [
+      [PARTICIPATED_IN, 0, entitySetName, studyEntityKeyId, CHRONICLE_STUDIES, {
+        [STATUS.toString()]: [ENROLLED]
+      }]
+    ];
+    let entityData = processEntityData(formData, entitySetIds, propertyTypeIds);
+    const associationEntityData = processAssociationEntityData(fromJS(associations), entitySetIds, propertyTypeIds);
```
UnsungHero97, line 518: Just fyi, `processAssociationEntityData` will work for both `associations` and `fromJS(associations)`.

src/containers/study/ParticipantsTable.js
```diff
@@ -0,0 +1,150 @@
...
+            components={components}
+            data={participants.valueSeq().toJS()}
+            headers={tableHeader}
+            paginated
+            rowsPerPageOptions={[5, 20, 50]} />
+      </TableWrapper>
+      <ParticipantInfoModal
+          handleOnClose={() => setInfoModalOpen(false)}
+          isVisible={infoModalOpen}
+          participantId={participants.getIn([participantEntityKeyId, PERSON_ID, 0])}
```
UnsungHero97, line 135: I recommend using the entity key id instead.

src/containers/study/ParticipantsTable.js
```diff
@@ -0,0 +1,150 @@
...
+          studyId={studyId} />
+      <DeleteParticipantModal
+          handleOnClose={() => setDeleteModalOpen(false)}
+          handleOnDeleteParticipant={() => handleOnDeleteParticipant()}
```
UnsungHero97: This can be just `handleOnDeleteParticipant={handleOnDeleteParticipant}`

src/containers/study/ParticipantsTable.js
```diff
@@ -0,0 +1,150 @@
...
+          isVisible={deleteModalOpen}
+          participantId={participants.getIn([participantEntityKeyId, PERSON_ID, 0])}
+          requestState={requestStates[DELETE_STUDY_PARTICIPANT]} />
+
+      <ChangeEnrollment
+          enrollmentStatus={participants.getIn([participantEntityKeyId, STATUS, 0])}
+          handleOnChangeEnrollment={() => handleOnChangeEnrollment()}
```
UnsungHero97: This can be just `handleOnChangeEnrollment={handleOnChangeEnrollment}`.

src/containers/study/components/ParticipantRow.js
```diff
@@ -0,0 +1,187 @@
...
+    enrollmentStatus,
+    icon,
+    onClickIcon,
+    participantEKId,
+    studyId
+  } = props;
+
+  let iconColor = NEUTRALS[0];
+  if (icon === faToggleOn && enrollmentStatus === ENROLLED) {
+    [, , iconColor] = PURPLES;
```
UnsungHero97: Why not `iconColor = PURPLES[2]`?

src/containers/study/ParticipantsTable.js
```diff
@@ -0,0 +1,150 @@
...
+    const { dataset } = currentTarget;
+    const { actionId } = dataset;
+    const { keyId } = dataset;
+
+    setParticipantEntityKeyId(keyId);
+
+    // actions
+    if (actionId === LINK) setInfoModalOpen(true);
+    if (actionId === DELETE) openDeleteModal();
+    if (actionId === TOGGLE_ENROLLMENT) openEnrollmentModel();
```
UnsungHero97: Use either a switch or if-else.

src/containers/study/StudyParticipants.js
```diff
@@ -40,30 +39,50 @@ const MissingParticipants = () => (
...
-        <AddParticipantModal
-            isVisible={isModalOpen}
-            onCloseModal={() => setModalOpen(false)}
-            study={study} />
-      </Card>
-    </Container>
+    <Card>
+      <CardSegment vertical>
+        <AddParticipantsButton
+            onClick={() => openAddParticipantModal()}
```
UnsungHero97:
```
onClick={openAddParticipantModal}
```

src/utils/constants/EnvTypes.js
```diff
@@ -0,0 +1,18 @@
+// @flow
+
+type EnvTypeEnum = {|
+  PRODUCTION:'production';
+  STAGING:'staging';
+  LOCAL:'local';
```
UnsungHero97: A space before `:`.

src/utils/constants/GlobalConstants.js
```diff
@@ -0,0 +1,9 @@
+// @flow
+
+const PARTICIPANTS_PREFIX = 'chronicle_participants_';
+const STUDY_LOGIN_ROOT = 'https://openlattice.com/chronicle/login?studyId=';
```
UnsungHero97: A constant should be a constant. Here, you're providing an incomplete constant and requiring special knowledge about this constant. It's weird that an incomplete part of the query string is defined here, and the other *required* part of the query string must be added elsewhere.

What's another approach you can take?

src/utils/constants/ParticipantActionTypes.js
```diff
@@ -0,0 +1,20 @@
+// @flow
+
+type ParticipantActionTypesEnum = {|
+  DELETE:'delete';
+  DOWNLOAD:'download';
+  LINK:'link';
+  TOGGLE_ENROLLMENT:'toggle_enrollment';
```
UnsungHero97: Spaces before `:`.

UnsungHero97 (changes requested):

src/containers/studies/StudiesSagas.js
anzioka: Ok, noted.

src/containers/study/components/ParticipantRow.js
anzioka: here's why:

Screen Shot 2020-01-10 at 2 07 02 PM

src/containers/study/ParticipantsTable.js
anzioka: Do you mean pass the entity key id as a prop to the ParticipantInfoModal component? I don't understand @UnsungHero97

src/containers/studies/StudiesSagas.js
anzioka: For each particular study we only need to fetch participants once (when the participants tab is selected). Subsequent selections shouldn't result in a new fetch since the data is already in redux store.

src/utils/constants/GlobalConstants.js
anzioka: Ok, perhaps it makes more sense to provide functions that return the completed string?

src/containers/study/components/ParticipantRow.js
UnsungHero97: I think we should ignore this warning for now. `[, , iconColor] = PURPLES;` is much less readable. Also, when the time comes to update LUK colors, it'll be much easier to replace/refactor `PURPLES[2]`.

src/containers/studies/StudiesSagas.js
anzioka: This looks much cleaner.

src/containers/study/ParticipantsTable.js
UnsungHero97: Yes, that's what I mean. However, after looking at it again, I see that you're actually using `participantId` for the URL, so you can ignore my comment.

src/containers/study/components/ParticipantRow.js
anzioka: I agree, the destructuring is weird. Will ignore the warning.

UnsungHero97 (approved):

PR #16, opened 2020-01-15
add custom form validation to require unique study participants

anzioka:
validation

### DIFF
--- a/src/containers/study/components/AddParticipantForm.js
+++ b/src/containers/study/components/AddParticipantForm.js
@@ -33,10 +33,9 @@ const AddParticipantForm = (props :Props, ref) => {
     dispatch(addStudyParticipant({ formData, studyEntityKeyId, studyId }));
   };
 
-  const validate = (formData, errors) => {
-    const updatedErrors = validateAddParticipantForm(formData, errors, participants, studyId);
-    return updatedErrors;
-  };
+  const validate = (formData, errors) => (
+    validateAddParticipantForm(formData, errors, participants, studyId)
+  );
 
   return (
     <Form
--- a/src/containers/study/components/AddParticipantForm.js
+++ b/src/containers/study/components/AddParticipantForm.js
@@ -33,6 +33,11 @@ const AddParticipantForm = (props :Props, ref) => {
     dispatch(addStudyParticipant({ formData, studyEntityKeyId, studyId }));
   };
 
+  const validate = (formData, errors) => {
+    const updatedErrors = validateAddParticipantForm(formData, errors, participants, studyId);
+    return updatedErrors;
+  };
+
   return (
     <Form
         hideSubmit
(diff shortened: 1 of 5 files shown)


src/containers/study/components/AddParticipantForm.js
```diff
@@ -38,7 +40,8 @@ const AddParticipantForm = (props :Props, ref) => {
         ref={ref}
         noPadding
         schema={dataSchema}
-        uiSchema={uiSchema} />
+        uiSchema={uiSchema}
+        validate={(formData, errors) => validateAddParticipantForm(formData, errors, participants, studyId)} />
```
UnsungHero97: I feel like there's some hook magic you can use here to avoid the inline function.

UnsungHero97 (changes requested):

src/containers/study/components/AddParticipantForm.js
anzioka: yeah you are probably right. Looking into it!

src/containers/study/components/AddParticipantForm.js
```diff
@@ -31,14 +33,20 @@ const AddParticipantForm = (props :Props, ref) => {
     dispatch(addStudyParticipant({ formData, studyEntityKeyId, studyId }));
   };
 
+  const validate = (formData, errors) => {
+    const updatedErrors = validateAddParticipantForm(formData, errors, participants, studyId);
+    return updatedErrors;
+  };
```
giuliacampana: Alternatively:
```
const validate = (formData, errors) => (
 validateAddParticipantForm(formData, errors, participants, studyId);
);

File: src/utils/FormUtils.js [hunk 3]
```diff
@@ -53,4 +62,44 @@ const createFormDataFromStudyEntity = (dataSchema :Object, study :Map) => {
...
+  return participantIds.includes(participantId.trim());
+};
+
+// custom react-jsonschema-form validation
+// ref: https://react-jsonschema-form.readthedocs.io/en/latest/validation/#custom-error-messages
+const validateAddParticipantForm = (formData :Object, errors :Object, participants :Map, studyId :UUID) => {
+  const participantsEntitySetName = getParticipantsEntitySetName(studyId);
+
+  const psk = getPageSectionKey(1, 1);
+  const eak = getEntityAddressKey(0, participantsEntitySetName, PERSON_ID);
```
giuliacampana: change isn't necessary—this is nitpicky, but I think it's better to use more descriptive variable names if something isn't actually an acronym.

giuliacampana (dismissed):

src/containers/study/components/AddParticipantForm.js
anzioka: Thanks! This looks cleaner.

src/utils/FormUtils.js
anzioka: You are right!

UnsungHero97 (approved):

PR #17, opened 2020-01-16
add search filter in the participants page

anzioka:
ezgif com-video-to-gif

search

### DIFF
--- a/src/containers/study/StudyParticipants.js
+++ b/src/containers/study/StudyParticipants.js
@@ -100,7 +100,7 @@ const StudyParticipants = ({ study } :Props) => {
     <Card>
       <CardSegment vertical>
         <CardHeader>
-          <SearchInputWrapper onChange={handleOnChange} />
+          <SearchInputWrapper onChange={handleOnChange} defaultValue="Filter..." />
           <AddParticipantsButton
               onClick={openAddParticipantModal}
               mode="primary">
--- a/src/containers/study/StudyParticipants.js
+++ b/src/containers/study/StudyParticipants.js
@@ -100,7 +100,7 @@ const StudyParticipants = ({ study } :Props) => {
     <Card>
       <CardSegment vertical>
         <CardHeader>
-          <SearchInputWrapper onChange={handleOnChange} defaultValue="Filter..." />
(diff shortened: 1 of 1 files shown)


anzioka:
Screen Shot 2020-01-16 at 2 43 13 PM

src/containers/study/StudyParticipants.js
```diff
@@ -53,15 +66,30 @@ const StudyParticipants = ({ study } :Props) => {
...
     dispatch(resetRequestState(ADD_PARTICIPANT));
     setModalOpen(true);
   };
 
+  const handleOnChange = (event :SyntheticInputEvent<HTMLInputElement>) => {
+    const { currentTarget } = event;
+    const { value } = currentTarget;
+
+    const matchingResults = participants
+      .valueSeq()
```
UnsungHero97: I don't think you need to do `.valueSeq()`.

src/containers/study/StudyParticipants.js
```diff
@@ -53,15 +66,30 @@ const StudyParticipants = ({ study } :Props) => {
     }
   }, [dispatch, participants, studyId]);
 
+  useEffect(() => {
+    setFilteredParticipants(participants);
+  }, [participants]);
```
UnsungHero97, line 67: Would this work?

```
const participants :Map = useSelector(...);
const [filteredParticipants, setFilteredParticipants] = useState(participants);
```

UnsungHero97 (changes requested):

src/containers/study/StudyParticipants.js
anzioka: You are right.

src/containers/study/StudyParticipants.js
anzioka: I tried that but it fails the first time the page is rendered (when `getStudyParticipants` is dispatched to fetch data). I think the reason is that `participants` is initially empty and therefore `filteredParticipants` is initialized as an empty map. After the data is fetched, `filteredParticipants` is not updated, thus the table is not rendered.

UnsungHero97 (approved):

PR #18, opened 2020-01-21
Task/set read write permissions

anzioka:
1. Grant chronicle super user READ and WRITE access on chronicle_participants_{uuid} entity sets.
2. Filter out studies that users have no READ access to.

### DIFF
--- a/src/containers/studies/StudiesSagas.js
+++ b/src/containers/studies/StudiesSagas.js
@@ -23,8 +23,10 @@ import {
   DataApiSagas,
   EntitySetsApiActions,
   EntitySetsApiSagas,
+  PermissionsApiActions,
+  PermissionsApiSagas,
   SearchApiActions,
-  SearchApiSagas
+  SearchApiSagas,
 } from 'lattice-sagas';
 import type { SequenceAction } from 'redux-reqseq';
 
@@ -46,12 +48,13 @@ import {
   getParticipantsEnrollmentStatus,
   getStudies,
   getStudyParticipants,
+  updateParticipantsEntitySetPermissions,
   updateStudy
 } from './StudiesActions';
 
 import EnrollmentStatuses from '../../utils/constants/EnrollmentStatus';
 import Logger from '../../utils/Logger';
-import { selectEntityTypeId } from '../../core/edm/EDMUtils';
+import { selectEntityType, selectEntityTypeId } from '../../core/edm/EDMUtils';
 import { ASSOCIATION_ENTITY_SET_NAMES, ENTITY_SET_NAMES } from '../../core/edm/constants/EntitySetNames';
 import { ENTITY_TYPE_FQNS, PROPERTY_TYPE_FQNS } from '../../core/edm/constants/FullyQualifiedNames';
 import { submitDataGraph, submitPartialReplace } from '../../core/sagas/data/DataActions';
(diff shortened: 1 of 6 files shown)


src/containers/studies/StudiesSagas.js
```diff
@@ -421,6 +453,66 @@ function* updateStudyWatcher() :Generator<*, *, *> {
...
+  const workerResponse = {};
+  try {
+    yield put(updateParticipantsEntitySetPermissions.request(action.id));
+
+    const participantsEntitySetId :UUID = action.value;
+
+    const entityType = yield select(selectEntityType(PERSON));
+    const updates :AclData[] = [];
+
+    entityType.toJS().properties.forEach((propertyTypeId :UUID) => {
```
UnsungHero97: Please avoid `.toJS()`.

src/core/edm/EDMUtils.js
```diff
@@ -8,6 +8,15 @@ const selectEntityTypeId = (entityTypeFQN :FQN) => (state :Map) => {
   return state.getIn(['edm', 'entityTypes', entityTypeIndex, 'id']);
 };
 
+const selectEntityType = (entityTypeFQN :FQN) => (state :Map) => {
+
+  const entityTypes = state
+    .getIn(['edm', 'entityTypes'])
+    .filter((entityType) => entityType.get('type').equals(Map(entityTypeFQN)));
+  return entityTypes.first();
```
UnsungHero97: You don't need to filter through all of the entity types, this is why `entityTypesIndexMap` exists:

```
const entityTypeIndex :number = state.getIn(['edm', 'entityTypesIndexMap', entityTypeFQN]);
return state.getIn(['edm', 'entityTypes', entityTypeIndex]);
```

src/containers/studies/StudiesSagas.js
```diff
@@ -552,6 +652,50 @@ function* addStudyParticipantWatcher() :Generator<*, *, *> {
...
+    const participantsEntitySetName = getParticipantsEntitySetName(studyId);
+
+    let response = yield call(getEntitySetIdWorker, getEntitySetId(participantsEntitySetName));
+    if (response.error) throw response.error;
+    const participantsEntitySetId = response.data;
+
+    const accessCheck = {
+      aclKey: [participantsEntitySetId],
+      permissions: [PermissionTypes.READ]
+    };
```
UnsungHero97: Please use`AccessCheckBuilder`.

src/containers/studies/StudiesSagas.js
```diff
@@ -572,8 +716,17 @@ function* getStudiesWorker(action :SequenceAction) :Generator<*, *, *> {
       throw response.error;
     }
 
+    const studyIds = response.data.map((study) => getIn(study, [STUDY_ID, 0]));
+    const readPermissions = yield all(
+      studyIds.map((studyId) => call(getStudyReadPermissionWorker, getStudyReadPermission(studyId)))
+    );
```
UnsungHero97: With this approach, you'll be making 1 access check request per study. You can do it in 1 request if you just pass all studies to `getStudyReadPermission` and then pass all access checks to `getAuthorizations`.

```
const studies = fromJS(response.data);
const studyAuthorizations = yield call(getStudyAuthorizationsWorker, getStudyAuthorizations({
  studies,
  permissions: [PermissionTypes.READ]
}));

/*
 * studyAuthorizations looks like this...
 * [{
 *   aclKey: ["the study id"],
 *   permissions: { READ: true } // or false
 * }]
 */

const authorizedStudyIds :Set<UUID> = Set().asImmutable((set :Set) => {
  studyAuthorizations.forEach((authorization) => {
    if (getIn(authorization, ['permissions', PermissionTypes.READ], false)) {
      set.add(getIn(authorization, ['aclKey', 0]));
    }
  });
});
```

Alternatively, you can move the `Set` logic and filter logic into the worker, rename it to something like `getAuthorizedStudies`, and then you get...


```
const studies = fromJS(response.data);
const authorizedStudies :Map<UUID, Map> = yield call(
  getAuthorizedStudiesWorker, 
  getAuthorizedStudies({ ... })
);
```

This way, all of the logic for an "authorized" study is contained in a single worker instead of multiple.

UnsungHero97 (changes requested):

UnsungHero97 (approved):

PR #23, opened 2020-02-13
add datetime property to participated_in associations

anzioka:
Add enrollment date for participants. When participants are unenrolled from a study, the datetime property is set to null and the corresponding row in the participants table has a blank entry for the date.

date

### DIFF
--- a/src/utils/DateUtils.js
+++ b/src/utils/DateUtils.js
@@ -1,0 +1,16 @@
+// @flow
+import { DateTime } from 'luxon';
+
+// @return 10/20/2020, 1:30 PM
+const getDateTimeFromIsoDate = (isoDate :string) => {
+  console.log('date', isoDate);
+  const date :DateTime = DateTime.fromISO(isoDate);
+  if (date.isValid) {
+    return date.toLocaleString(DateTime.DATETIME_SHORT);
+  }
+  return null;
+};
+
+export {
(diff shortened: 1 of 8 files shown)


src/utils/DateUtils.js
```diff
@@ -0,0 +1,16 @@
+// @flow
+import { DateTime } from 'luxon';
+
+// @return 10/20/2020, 1:30 PM
```
solomontang, line 4: This comment is only true if the consuming browser locale matches 'en-US'. This means free Intl! Far superior than explicitly defining a format like 'MM/DD/YYYY, hh:mm a'.

> https://moment.github.io/luxon/docs/class/src/datetime.js~DateTime.html#instance-method-toLocaleString
Returns a localized string representing this date. Accepts the same options as the Intl.DateTimeFormat constructor and any presets defined by Luxon, such as DateTime.DATE_FULL or DateTime.TIME_SIMPLE. The exact behavior of this method is browser-specific, but in general it will return an appropriate representation of the DateTime in the assigned locale. Defaults to the system's locale if no locale has been specified

solomontang (approved):
LGTM

src/utils/DateUtils.js
anzioka: Good point!

PR #24, opened 2020-02-15
Feature/apps 1739 create chronicle survey

anzioka:
Display apps used by participants filtered by the current date
survey

**Successful submission:**
successful

**Submission failure:**
error_state

**Mobile View:**
mobile


package.json
```diff
@@ -32,6 +32,7 @@
     "@fortawesome/pro-solid-svg-icons": "~5.11.0",
     "@fortawesome/react-fontawesome": "~0.1.0",
     "@redux-saga/core": "~1.1.0",
+    "axios": "~0.19.2",
```
UnsungHero97: Please keep patch version values at 0.

package.json
```diff
@@ -44,6 +45,7 @@
     "loglevel": "~1.6.0",
     "luxon": "~1.21.0",
     "polished": "~3.4.0",
+    "qs": "~6.9.1",
```
UnsungHero97: Please keep patch version values at 0.

src/utils/constants/AppUserTypes.js
```diff
@@ -0,0 +1,18 @@
+// @flow
+
+type AppUserTypeEnum = {|
+  PARENT :'parent';
+  CHILD :'child';
+  PARENT_AND_CHILD :'parent_and_child';
+|};
+
+type AppUserType = $Values<AppUserTypeEnum>;
```
UnsungHero97: I found a slightly better way to do this:

```
const AppUserTypes :{|...AppUserTypeEnum|} = Object.freeze({
 ...
});
type AppUserType = $Values<typeof AppUserTypes>;
```

src/utils/api/AppApi.js
```diff
@@ -13,6 +13,7 @@ import {
   STUDY,
 } from '../constants/UrlConstants';
 import type { ParticipantDataType } from '../constants/ParticipantDataTypes';
+import ParticipantDataTypes from '../constants/ParticipantDataTypes';
```
UnsungHero97: Please run the import sorter.

src/index.js
```diff
@@ -16,6 +17,8 @@ import AppContainer from './containers/app/AppContainer';
 import initializeReduxStore from './core/redux/ReduxStore';
 import initializeRouterHistory from './core/router/RouterHistory';
 import * as Routes from './core/router/Routes';
+import SurveyContainer from './containers/survey/SurveyContainer';
```
UnsungHero97: Please run the import sorter.

src/containers/survey/SurveyContainer.js
```diff
@@ -0,0 +1,180 @@
...
+  font-size: 16px;
+  font-weight: 400;
+  margin: 5px 0 20px 0;
+`;
+
+const SurveyContainer = () => {
+  const dispatch = useDispatch();
+  const history = useHistory();
+
+  const queryParams = qs.parse(history.location.search, { ignoreQueryPrefix: true });
```
UnsungHero97: Question: are the params before or after the hash `#`?

If before, I don't see a reason to use `history` from `react-router`; just use `window.location.search`. If after, you should use `location` instead of `history` via `useLocation()`.

src/containers/survey/SurveyContainer.js
```diff
@@ -0,0 +1,180 @@
...
+        <Spinner size="2x" />
+      </SpinnerWrapper>
+    );
+  }
+
+  const ErrorMessage = () => (
+    <ErrorWrapper>
+        Sorry, something went wrong. Please try refreshing the page, or contact support.
+    </ErrorWrapper>
+  );
```
UnsungHero97: This doesn't need to be inside `SurveyContainer`. It can safely be defined outside, it's basically a constant.

src/containers/survey/SurveyContainer.js
```diff
@@ -0,0 +1,180 @@
...
+                {
+                  requestStates[SUBMIT_SURVEY] === RequestStates.SUCCESS
+                    ? <SubmissionSuccessful />
+                    : (
+                      <>
+                        <SurveyTitle>
+                          Apps Usage Survey
+                        </SurveyTitle>
+                        <CurrentDate>
+                          {getCurrentDate()}
```
UnsungHero97: This doesn't need to be a function.

src/containers/survey/SurveyContainer.js
```diff
@@ -0,0 +1,180 @@
...
+
+  const { appsData, requestStates } = useSelector((state) => ({
+    appsData: state.getIn(['appsData', 'appsData'], Map()),
+    requestStates: {
+      [GET_CHRONICLE_APPS_DATA]: state.getIn(['appsData', GET_CHRONICLE_APPS_DATA, 'requestState']),
+      [SUBMIT_SURVEY]: state.getIn(['appsData', SUBMIT_SURVEY, 'requestState'])
+    }
+  }));
+
+  const getCurrentDate = () => getFullDateFromIsoDate(new Date().toISOString());
```
UnsungHero97: It looks like `getFullDateFromIsoDate` is only used here and only used with a new date. Based on that, `getFullDateFromIsoDate` doesn't need to exist because you can accomplish the same thing like this:

```
const fullDate :string = DateTime.local().toLocaleString(DateTime.DATE_FULL); // 'February 25, 2020'
```

src/containers/survey/SurveyReducer.js
```diff
@@ -0,0 +1,57 @@
...
+import {
+  GET_CHRONICLE_APPS_DATA,
+  SUBMIT_SURVEY,
+  getChronicleAppsData,
+  submitSurvey
+} from './SurveyActions';
+
+import { RESET_REQUEST_STATE } from '../../core/redux/ReduxActions';
+
+const INITIAL_STATE :Map<*, *, *> = fromJS({
```
UnsungHero97: The flow annotation is wrong `:Map<*, *, *>`. Maps only have a key and a value. Separately, since we're not yet using immutable's flow definitions, drop the `*` and just use `:Map`.

src/containers/survey/SurveyTable.js
```diff
@@ -0,0 +1,155 @@
...
+    <SubmissionFailureModal
+        handleOnClose={hideErrorModal}
+        isVisible={errorModalVisible} />
+  );
+
+  const components = {
+    Row: ({ data: rowData } :any) => (
+      <TableRow data={rowData} handleOnChange={handleOnChange} />
+    )
+  };
```
UnsungHero97: See @solomontang's comment on this: https://github.com/Lattice-Works/pretrial-case-management/pull/728#discussion_r383566077

src/containers/survey/SurveyTable.js
```diff
@@ -0,0 +1,155 @@
...
+
+                <SubmitButtonWrapper>
+                  <Button
+                      isLoading={submitRequestState === RequestStates.PENDING}
+                      mode="primary"
+                      onClick={handleOnSubmit}>
+                      Submit Survey
+                  </Button>
+                </SubmitButtonWrapper>
+                { renderErrorModal() }
```
UnsungHero97: I don't see a reason to have `SubmissionFailureModal` inside the function. Why not just place it here directly?

src/containers/survey/components/TableRow.js
```diff
@@ -0,0 +1,88 @@
...
+  handleOnChange :(SyntheticInputEvent<HTMLInputElement>) => void;
+};
+
+const TableRow = ({ data, handleOnChange } :Props) => {
+  const appName :string = getIn(data, ['entityDetails', TITLE]);
+  const appEntityId :UUID = get(data, 'id');
+  const appUsers :Set = getIn(data, ['associationDetails', USER_FQN], Set());
+
+  return (
+    <>
```
UnsungHero97: `RowWrapper` wraps everything, so you don't need the fragments.

UnsungHero97 (changes requested):

src/containers/survey/SurveySagas.js
```diff
@@ -0,0 +1,153 @@
...
+function* submitSurveyWorker(action :SequenceAction) :Generator<*, *, *> {
+  try {
+    yield put(submitSurvey.request(action.id));
+
+    const { value } = action;
+    const { participantId, studyId, appsData } = value;
+
+    const url = getParticipantUserAppsUrl(participantId, studyId);
+
+    if (url === null) throw new Error('Invalid Url');
```
UnsungHero97: `if (!url)` covers `null` and other falsy values.

src/containers/survey/SurveySagas.js
```diff
@@ -0,0 +1,153 @@
...
+
+    let associationData :Map = Map().withMutations((map) => {
+      appsData.forEach((entry, key) => {
+        map.set(key, entry.get('associationDetails'));
+      });
+    });
+
+    // delete entity key since it won't be used in DataApi.updateEntitiesInEntitySet()
+    associationData = associationData
+      .map((entity) => entity.delete(OPENLATTICE_ID_FQN));
```
UnsungHero97: You can avoid this reassignment by moving the `entity.delete` into the `forEach` above.

src/containers/survey/SurveySagas.js
```diff
@@ -0,0 +1,153 @@
...
+        }
+     */
+
+    const axiosRequestBody :Object = associationData.toJS();
+
+    const response = yield call(axios, {
+      method: 'post',
+      data: axiosRequestBody,
+      url,
+    });
```
UnsungHero97: This block of code belongs in something like `ChronicleApi`, similar to `lattice-js`. Please refactor:

```
const response = yield call(ChronicleApi.updateAppsUsageAssociationData, data);
```

src/containers/survey/SurveySagas.js
```diff
@@ -0,0 +1,153 @@
...
+            }
+          ]
+     */
+    const url = getParticipantUserAppsUrl(participantId, studyId);
+    if (url === null) throw new Error('Invalid url');
+
+    const response = yield call(axios, {
+      method: 'get',
+      url,
+    });
```
UnsungHero97: This block of code belongs in something like `ChronicleApi`, similar to `lattice-js`. Please refactor:

```
const response = yield call(ChronicleApi.getParticipantAppsUsageData);
```

UnsungHero97 (changes requested):

anzioka:
after replacing useState with useReducer: 
survey

UnsungHero97 (approved):

PR #84, opened 2020-05-21
filter out studies whose access to notification entity sets is unauthorized

anzioka:
PR: 
Only display studies whose `getEntitySetId` on the corresponding notifications / part_of entity sets does not result in 401 status code.


anzioka:
No longer needed.

PR #120, opened 2020-06-11
update participant data endpoint

anzioka:
PR: add /authenticated to participant data url


UnsungHero97 (approved):
Let's wait until `chronicle-server` is released.

PR #123, opened 2020-06-15
Task/apps 2070 create questionnaires

anzioka:
PR:

1) Questionnaires dashboard. Display both active & inactive questionnaires in a table. Filter by active/inactive

2) Delete questionnaires and associated questions 

3) Toggle active status

4) Create questionnaire

dash

delete

create

ezgif com-video-to-gif


package.json
```diff
@@ -39,9 +39,9 @@
     "immutable": "4.0.0-rc.9",
     "lattice": "0.73.0",
     "lattice-auth": "0.18.0",
-    "lattice-fabricate": "0.13.1",
+    "lattice-fabricate": "^0.13.1",
```
UnsungHero97: Please use `~`.

package.json
```diff
@@ -39,9 +39,9 @@
...
     "lattice-sagas": "0.47.3",
-    "lattice-ui-kit": "0.32.0",
+    "lattice-ui-kit": "^0.32.0",
```
UnsungHero97: Please use `~`.

UnsungHero97 (changes requested):

UnsungHero97 (changes requested):
Please capitalize the file names.

solomontang:
Any chance you could show us what the created questionnaire looks like? Also I'm noticing there isn't an "Edit" functionality. Is that coming in a future PR?

src/containers/questionnaires/QuestionnairesContainer.js
```diff
@@ -0,0 +1,195 @@
...
+  // initially display all questionnaires
+  useEffect(() => {
+    setSelectedStatus(STATUS_SELECT_OPTIONS);
+  }, []);
+
+  useEffect(() => {
+    if (studyQuestionnaires.isEmpty()) {
+      dispatch((getStudyQuestionnaires(studyEKID)));
+    }
+  }, [studyQuestionnaires, studyEKID, dispatch]);
```
solomontang, line 86: Will this run into an infinite loop of requests if getStudyQuestionnaires doesn't return any questionnaires? For example, what happens if there is a new study that doesn't have any questionnaires yet? Or if someone deletes a questionnaire to attempt to replace it?

src/containers/questionnaires/components/CreateQuestionnaireForm.js
```diff
@@ -0,0 +1,202 @@
...
+
+const ActionRow = styled.div`
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  margin-top: 20px;
+`;
+
+type Props = {
+  onExitEditMode :() => void;
```
solomontang: From what I can tell, this is basically an onClose function. I'm not sure what "edit mode" is as it doesn't seem like you can edit existing questionnaires (only create, delete, and inactivate). If this is true, can you rename this prop?

src/containers/questionnaires/components/CreateQuestionnaireForm.js
```diff
@@ -0,0 +1,202 @@
...
+      onExitEditMode();
+    }
+    dispatch(resetRequestState(CREATE_QUESTIONNAIRE));
+  };
+
+  return (
+    <CardSegment padding="20px">
+      <Header>
+        <h3> Create Questionnaire </h3>
+        <Button onClick={onExitEditMode}> Cancel </Button>
```
solomontang: Yeah, looks like invoking `onExitEditMode` is intended to just cancel creating a new questionnaire. Please rename to `onClose` or something that better represents what the function is doing.

anzioka:
Here is a sample questionnaire:
Screen Shot 2020-07-02 at 3 18 46 PM

src/containers/questionnaires/components/QuestionnairePreview.js
```diff
@@ -0,0 +1,70 @@
...
+  title
+} :Props) => {
+  const questionEntities = createPreviewQuestionEntities(formData);
+  const schemaProperties = getSchemaProperties(questionEntities);
+
+  const uiSchemaOptions = getUiSchemaOptions(schemaProperties);
+  const { schema, uiSchema } = createSchema(schemaProperties, uiSchemaOptions);
+  return (
+    <Modal isVisible={isVisible} onClose={onClose} textTitle="Questionnaire Preview">
+      <div style={{ paddingBottom: '20px' }}>
```
solomontang: I get that inline styles are convenient especially for small changes like this, but it's prob best to stick to styled-components. Kind of strange to see some elements are using it but some are not. A little harder to see where some of the inline styles end since they're all just generic div tags as well.

A nitpick for sure.

src/containers/questionnaires/schemas/questionnaireSchema.js
```diff
@@ -0,0 +1,240 @@
...
+  [getPageSectionKey(2, 1)]: {
+    classNames: 'column-span-12',
+    'ui:options': {
+      addButtonText: '+ Add Question'
+    },
+    items: {
+      classNames: 'grid-container',
+      [getEntityAddressKey(-1, QUESTIONS_ES_NAME, TITLE_FQN)]: {
+        classNames: 'column-span-12',
+        'ui:autofocus': true
```
solomontang, line 145: I'd be careful with adding autofocus. Can have some intended behaviors when you're going back and forth between pages especially on mobile. Super annoying on mobile to have the on screen keyboard pop up all the time. Especially tricky since this is being on a list of items so it's possible for multiple elements per page to fight for focus.

src/containers/questionnaires/table/TableHeaderRow.js
```diff
@@ -0,0 +1,35 @@
...
+  className,
+  headers
+} :Props) => (
+  <thead className={className}>
+    <StyledRow sticky>
+      {
+        headers.map((header) => (
+          <components.HeadCell
+              key={header.key}
+              width={header.key} />
```
solomontang: I suspect you mean to set the width to something other than the key which i believe it set to the following from https://github.com/Lattice-Works/chronicle/pull/123/files#diff-1af0a8e748242401224a609ed34e4cb0R39

```
const tableHeaders = ['title', 'status', 'actions'].map((header) => ({
  key: header,
  label: '',
  sortable: false
}));
```

src/containers/questionnaires/utils/index.js
```diff
@@ -0,0 +1,184 @@
...
+
+  return result;
+};
+
+// return [hour, minute] from a 24 hour formated time string
+// input example: '21:00', '09:00'
+const parseTimeStr = (time :string) => {
+  const tokens = time.split(':');
+  return tokens.map((token) => parseInt(token, 10));
+};
```
solomontang: There's some functionality built into luxon to do this for you. From any valid DateTime, you can access a series of properties like `.hour` and `.minute`

DateTime.local().hour

solomontang (changes requested):

src/containers/questionnaires/QuestionnairesContainer.js
anzioka: No it won't run into an infinite loop. If the request doesn't return any data, then  studyQuestionnaires is just an empty immutable map and won't trigger subsequent renders + requests. Here is the view for a study that has no questionnaires.

Screen Shot 2020-07-02 at 5 16 14 PM

src/containers/questionnaires/components/QuestionnairePreview.js
anzioka: Good point. Consistency

src/containers/questionnaires/schemas/questionnaireSchema.js
anzioka: ah I hadn't thought about how this could impact user experience on mobile. I think the upside of autofocus is that it makes text inputting so much easier (at least on large screens). With autofocus, adding a new question is one click simpler. As for multiple views competing for focus, it seems to me, after several attempts, that the last view in the list will acquire focus on going back to previous page. That doesn't invalidate your point though haha.

It makes sense for this tool to be used primarily on large devices though, so maybe we could shrug off potential bad experience on mobile?

src/containers/questionnaires/table/TableHeaderRow.js
anzioka: I actually intent to set the width to the key.

The keys are mapped to widths in the getStyleVariation util:

https://github.com/Lattice-Works/chronicle/blob/e73748c41a19ac4f9171a812d40a06e2afead137/src/containers/questionnaires/QuestionnairesContainer.js#L63

```
const getWidthVariation = getStyleVariation('width', {
  default: 'auto',
  title: 'auto',
  status: '100px',
  actions: '110px'
});

const Cell = styled.td`
  width: ${getWidthVariation};
`;

type HeadCellProps = {
  width :string;
};

const HeadCell = ({ width } :HeadCellProps) => (
  <Cell width={width} />
);

```

src/containers/questionnaires/components/CreateQuestionnaireForm.js
anzioka: renamed!

anzioka:
> Any chance you could show us what the created questionnaire looks like? Also I'm noticing there isn't an "Edit" functionality. Is that coming in a future PR?

Edit functionality is on its way in a separate PR :)

src/containers/questionnaires/QuestionnairesContainer.js
```diff
@@ -0,0 +1,195 @@
...
+                {
+                  tableData.length === 0 ? (
+                    <div style={{ textAlign: 'center', padding: '60px' }}>
+                      No questionnaires found.
+                    </div>
+                  ) : (
+                    <Table
+                        components={{ HeadCell, Header: TableHeaderRow, Row: RowComponent }}
+                        data={tableData}
+                        headers={tableHeaders} />
```
UnsungHero97: Why did you go with `Table` here? This approach seems overly complicated and unnecessary.

1. the `Table` background does not respect the `Card` border radius:
Screen Shot 2020-07-08 at 10 13 33 AM

2. the `Table` does not respect the `CardSegment` padding:
Screen Shot 2020-07-08 at 10 05 28 AM

3. the `Table` header doesn't seem to be used, so there's unnecessary elements in the DOM that do nothing yet require computation from React.
Screen Shot 2020-07-08 at 10 10 54 AM

src/containers/questionnaire/QuestionnaireSagas.js
```diff
@@ -80,6 +108,215 @@ const {
...
+
+    const questionnaireESID = yield select(selectEntitySetId(QUESTIONNAIRE_ES_NAME));
+    const activePTID = yield select(selectPropertyTypeId(ACTIVE_FQN));
+
+    const response = yield call(updateEntityDataWorker, updateEntityData({
+      entitySetId: questionnaireESID,
+      entities: {
+        [questionnaireEKID]: {
+          [activePTID]: [!activeStatus]
+        }
```
UnsungHero97: I think this is a pattern that should be avoided as it can easily lead to preventable bugs.

What `changeActiveStatus` is REALLY doing is actually just setting the property value, i.e. `setActiveStatus`. As it is currently implemented, the actual logic for determining the value lives in 2 different places - `!activeStatus` is in the worker saga and in the reducer. 

For example, if `activeStatus` was a number instead of a boolean, you wouldn't want to do `activeStatus + 10` in multiple places. Or, if the logic for determining the `activeStatus` value was more complex, you wouldn't want to implement that in multiple places.

Further, the value for this particular property is strongly coupled with the user action. In other words, the value of `activeStatus` is directly controlled by the user's action. As such, I think `!activeStatus` should near the user action.

```
dispatch(
  setActiveStatus({
    questionnaireEKID,
    studyEKID,
    activeStatus: !activeStatus,
  })
);
```

src/containers/questionnaire/QuestionnaireSagas.js
```diff
@@ -80,6 +108,215 @@ const {
...
+    const neighborFilter = {
+      entityKeyIds: [questionnaireEKID],
+      destinationEntitySetIds: [],
+      sourceEntitySetIds: [questionsESID]
+    };
+
+    const response = yield call(deleteEntityAndNeighborDataWorker, deleteEntityAndNeighborData({
+      entitySetId: questionnaireESID,
+      filter: neighborFilter,
+      deleteType: DeleteTypes.HARD
```
UnsungHero97, line 179: My understanding is that `DeleteTypes.HARD` is for special cases. Let's discuss `SOFT` vs `HARD`.

src/containers/questionnaire/QuestionnaireSagas.js
```diff
@@ -80,6 +108,215 @@ const {
...
+  finally {
+    yield put(deleteQuestionnaire.finally(action.id));
+  }
+}
+
+function* deleteQuestionnaireWatcher() :Saga<*> {
+  yield takeEvery(DELETE_QUESTIONNAIRE, deleteQuestionnaireWorker);
+}
+
+function constructEntitiesFromFormData(formData, entityKeyIds, entitySetIds, propertyTypeIds :Map) {
```
UnsungHero97: This belongs in a `Utils` file, just like `createRecurrenceRuleSetFromFormData`.

src/core/edm/constants/FullyQualifiedNames.js
```diff
@@ -51,7 +51,10 @@ const PROPERTY_TYPE_FQNS = {
   DESCRIPTION_FQN: FQN.of('ol.description'),
   VALUES_FQN: FQN.of('ol.values'),
   DATE_TIME_FQN: FQN.of('ol.datetime'),
-  TITLE_FQN: FQN.of('ol.title')
+  TITLE_FQN: FQN.of('ol.title'),
+  RRULE_FQN: FQN.of('ol.rrule'),
+  ACTIVE_FQN: FQN.of('ol.active'),
+  COMPLETED_DATE_TIME_FQN: FQN.of('date.completeddatetime')
```
UnsungHero97: Please sort.

src/containers/questionnaire/QuestionnaireSagas.js
```diff
@@ -80,6 +108,215 @@ const {
...
+    const rruleSet = createRecurrenceRuleSetFromFormData(formData);
+
+    // update formdata with rrule
+    let psk = getPageSectionKey(1, 1);
+    let eak = getEntityAddressKey(0, QUESTIONNAIRE_ES_NAME, RRULE_FQN);
+    formData = setIn(formData, [psk, eak], rruleSet);
+
+    // set ol.active to true
+    eak = getEntityAddressKey(0, QUESTIONNAIRE_ES_NAME, ACTIVE_FQN);
+    formData = setIn(formData, [psk, eak], true);
```
UnsungHero97: This belongs in the schema as a default value, and you won't need these 2 lines.

```
[getEntityAddressKey(0, QUESTIONNAIRE_ES_NAME, ACTIVE_FQN)]: {
  default: true,
  type: 'boolean',
}
```

src/containers/questionnaire/QuestionnaireSagas.js
```diff
@@ -80,6 +108,215 @@ const {
...
+    yield put(createQuestionnaire.success(action.id, {
+      questionEntities,
+      questionnaireEntity,
+      studyEKID
+    }));
+
+  }
+  catch (error) {
+    yield put(createQuestionnaire.failure(action.id));
+    LOG.error(action.type, error);
```
UnsungHero97, line 269: `LOG` before `yield`.

src/containers/questionnaire/QuestionnaireSagas.js
```diff
@@ -180,9 +417,10 @@ function* getStudyQuestionnairesWorker(action :SequenceAction) :Saga<*> {
     const studyQuestionnaires = Map().withMutations((mutator) => {
       fromJS(response.data).get(studyEKID, List()).forEach((neighbor) => {
         const neighborId = neighbor.get('neighborId');
-        const details = neighbor.get('neighborDetails');
+        const details = neighbor.get('neighborDetails').asMutable();
+        details.set('id', neighborId); // needed by LUK table
 
-        mutator.set(neighborId, details);
+        mutator.set(neighborId, details.asImmutable());
```
UnsungHero97: You don't need to do the `asMutable()` / `asImmutable` dance.

src/containers/questionnaires/components/CreateQuestionnaireForm.js
```diff
@@ -0,0 +1,202 @@
...
+
+  const dispatch = useDispatch();
+
+  const [submitStatusModalTitle, setSubmitStatusModalTitle] = useState('');
+  const [submitStatusModalMessage, setSubmitStatusModalMessage] = useState('');
+  const [submitStatusModalVisible, setSubmitStatusModalVisible] = useState(false);
+
+  const requestStates = useSelector((state) => ({
+    [CREATE_QUESTIONNAIRE]: state.getIn(['questionnaire', CREATE_QUESTIONNAIRE, REQUEST_STATE])
+  }));
```
UnsungHero97: Don't return an object from `useSelector`.

```
const createQuestionnaireRS :?RequestState = useSelector(
  (state) => state.getIn(['questionnaire', CREATE_QUESTIONNAIRE, REQUEST_STATE])
);
```

Alternatively, you can use the `lattice-utils` helper:

```
import { useRequestState } from 'lattice-utils';
const createQuestionnaireRS :?RequestState = useRequestState(['questionnaire', CREATE_QUESTIONNAIRE]);
```

src/containers/questionnaires/components/DeleteQuestionnaireModal.js
```diff
@@ -0,0 +1,78 @@
...
+        onClose={onClose}
+        requestState={deleteQuestionnaireRS}
+        requestStateComponents={requestStateComponents}
+        textPrimary="Delete"
+        textSecondary="Cancel"
+        textTitle="Delete Questionnaire" />
+  );
+};
+
+export default React.memo<Props>(DeleteQuestionnaireModal);
```
UnsungHero97: Why `React.memo`?

src/containers/questionnaires/constants/constants.js
```diff
@@ -0,0 +1,57 @@
...
+// days of week
+const DAYS_OF_WEEK = {
+  MO: 'Monday',
+  TU: 'Tuesday',
+  WE: 'Wednesday',
+  TH: 'Thursday',
+  FR: 'Friday',
+  SA: 'Saturday',
+  SU: 'Sunday',
+};
```
UnsungHero97: Unless there's some special logic for this, I recommend using luxon directly:

https://moment.github.io/luxon/docs/class/src/info.js~Info.html#static-method-weekdays

src/containers/questionnaires/constants/questionTypes.js
```diff
@@ -0,0 +1,17 @@
+// @flow
+
+// questionnaire question types
+type QuestionTypeEnum = {|
+  TEXT_ENTRY:'Text entry';
+  MULTIPLE_CHOICE:'Multiple choice';
```
UnsungHero97: Formatting... space before `:`.

UnsungHero97: Please sort.

src/containers/questionnaires/constants/questionTypes.js
```diff
@@ -0,0 +1,17 @@
...
+|};
+
+type QuestionType = $Values<QuestionTypeEnum>;
+
+const QuestionTypes :{|...QuestionTypeEnum |} = Object.freeze({
+  TEXT_ENTRY: 'Text entry',
+  MULTIPLE_CHOICE: 'Multiple choice'
```
UnsungHero97: Please sort.

src/containers/questionnaires/constants/questionnaireStatus.js
```diff
@@ -0,0 +1,17 @@
+// @flow
+
+// questionnaire status
+type QuestionnaireStatusEnum = {|
+  ACTIVE:'ACTIVE';
+  NOT_ACTIVE:'NOT_ACTIVE';
```
UnsungHero97: Formatting... space before `:`

src/containers/questionnaires/constants/constants.js
UnsungHero97: Ok, I think I have a simpler alternative...

In the `schedulerSchema`, use `enum` for the value and `enumNames` for the label:

```
import { Info } from 'luxon';
items: {
  enum: [RRule.MO, RRule.TU, RRule.WE, RRule.TH, RRule.FR, RRule.SA, RRule.SU],
  enumNames: Info.weekdays(), // default order is Monday - Sunday
  type: 'string',
},
```

I'm pretty sure doing this will remove the need for the `DAYS_OF_WEEK` object as well as the `getRRuleWeekDayConstants` helper. 

You'll also no longer need `Object.entries(DAYS_OF_WEEK)` in `getWeekDaysAndTimesFromRruleSet` because you can use `Info.weekdays()` directly.

https://moment.github.io/luxon/docs/class/src/info.js~Info.html#static-method-weekdays
https://react-jsonschema-form.readthedocs.io/en/latest/usage/single/#custom-labels-for-enum-fields

src/containers/questionnaires/schemas/questionnaireSchema.js
```diff
@@ -0,0 +1,240 @@
...
+      type: 'object',
+      title: 'Notification trigger settings',
+      properties: {
+        daysOfWeek: {
+          title: 'Days of week to send notification',
+          type: 'array',
+          items: {
+            type: 'string',
+            enum: Object.values(DAYS_OF_WEEK)
+          },
```
UnsungHero97: Refer to my comment above.

src/core/sagas/Sagas.js
```diff
@@ -55,11 +56,17 @@ export default function* sagas() :Generator<*, *, *> {
     fork(SurveySagas.getChronicleUserAppsWatcher),
     fork(SurveySagas.submitSurveyWatcher),
 
+    // // questionnaires
+    // fork(QuestionnairesSagas.createQuestionnaireWatcher),
```
UnsungHero97: Please remove.

src/core/redux/ReduxReducer.js
```diff
@@ -22,6 +23,7 @@ export default function reduxReducer(routerHistory :any) {
     edm: edmReducer,
     permissions: permissionsReducer,
     questionnaire: questionnareReducer,
+    // questionnaires: questionnairesReducer,
```
UnsungHero97: Any reason to keep this around?

src/containers/questionnaires/utils/index.js
```diff
@@ -0,0 +1,177 @@
...
+  const associations = questions.map((question :Object, index :number) => [
+    PART_OF_ES_NAME, index, QUESTIONS_ES_NAME, 0, QUESTIONNAIRE_ES_NAME, {
+      [ID_FQN.toString()]: [uuid()]
+    }
+  ]);
+
+  // $FlowFixMe
+  return associations.concat(
+    [[PART_OF_ES_NAME, 0, QUESTIONNAIRE_ES_NAME, studyEKID, CHRONICLE_STUDIES, {
+      [COMPLETED_DATE_TIME_FQN.toString()]: [new Date()]
```
UnsungHero97, line 168: Shouldn't this be `new Date().toISOString()`?

UnsungHero97 (changes requested):

src/core/redux/ReduxReducer.js
anzioka: none. removed!

src/containers/questionnaires/utils/index.js
anzioka: I don't think it's necessary. The written value is a valid ISO date
Screen Shot 2020-07-08 at 5 51 43 PM

src/containers/questionnaires/components/DeleteQuestionnaireModal.js
anzioka: Not necessary :)

src/containers/questionnaires/utils/index.js
UnsungHero97: 🤯

UnsungHero97: Wait, I think you're looking at the wrong entity. The entity in the screenshot does not look like it's the `ol.partof` association entity because "Version" is not a property of `ol.partof`.

src/containers/questionnaires/constants/constants.js
anzioka: Awesome :)

anzioka: Except that the property type in the schema should be an object otherwise validation test fails

UnsungHero97: What do you mean?

anzioka: The enums are not strings. For example: RRule.MO
Screen Shot 2020-07-08 at 6 51 34 PM

In any case, when I used type 'string' I got a validation error.
Screen Shot 2020-07-08 at 6 53 29 PM

I think rjsf validates against the enum values, not enumNames

src/containers/questionnaires/utils/index.js
anzioka: It's actually the right entity. "Version" is part of the dst entity in the ol.partof association.

here is a larger screenshot
Screen Shot 2020-07-08 at 6 58 38 PM

src/containers/questionnaire/QuestionnaireSagas.js
anzioka: Makes sense!

src/containers/questionnaires/components/CreateQuestionnaireForm.js
```diff
@@ -0,0 +1,200 @@
...
+  [ABOUT_PAGE]: aboutSchema,
+  [QUESTIONS_PAGE]: questionsSchema,
+  [SCHEDULER_PAGE]: schedulerSchema
+};
+
+const Header = styled.div`
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  margin-bottom: 20px;
```
giuliacampana: please alphabetize :)

src/containers/questionnaires/components/CreateQuestionnaireForm.js
```diff
@@ -0,0 +1,200 @@
...
+    padding: 0;
+    margin: 0;
+  }
+`;
+
+const ActionRow = styled.div`
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  margin-top: 20px;
```
giuliacampana: please alphabetize :)

src/containers/questionnaires/components/NewQuestionnaireConfirmation.js
```diff
@@ -0,0 +1,105 @@
...
+const {
+  DESCRIPTION,
+  NUM_MULTIPLE_CHOICE,
+  NUM_SINGLE_ANSWER,
+  TITLE,
+} = QUESTIONNAIRE_SUMMARY;
+
+const StyledButton = styled(Button)`
+  margin-top: 20px;
+  align-self: flex-start;
```
giuliacampana: please alphabetize :)

src/containers/questionnaires/components/NewQuestionnaireConfirmation.js
```diff
@@ -0,0 +1,105 @@
...
+      <Wrapper>
+        <Badge count={get(summary, NUM_MULTIPLE_CHOICE)} />
+        <Label>
+          &nbsp; Multiple choice questions
+        </Label>
+      </Wrapper>
+
+      <Wrapper>
+        <Badge count={get(summary, NUM_SINGLE_ANSWER)} />
+        <Label>
```
giuliacampana, line 86: do you think it's necessary to pass a default value to any of these lookups in `summary`?

src/containers/questionnaires/components/QuestionnaireDetailsModal.js
```diff
@@ -0,0 +1,147 @@
...
+const HorizontalList = styled.ul`
+  padding: 0;
+
+  > li {
+    display: inline;
+    padding: 12px 20px;
+    font-size: 14px;
+    background-color: ${NEUTRALS[5]};
+    color: ${NEUTRALS[0]};
+    margin-right: 10px;
```
giuliacampana: please alphabetize :)

src/containers/questionnaires/components/QuestionnaireDetailsModal.js
```diff
@@ -0,0 +1,147 @@
...
+  }
+`;
+
+const SectionPane = styled.div`
+  border-bottom: 1px solid ${NEUTRALS[4]};
+  padding-bottom: 30px;
+  margin-bottom: 30px;
```
giuliacampana: please alphabetize :)

src/containers/questionnaires/components/QuestionnaireDetailsModal.js
```diff
@@ -0,0 +1,147 @@
...
+  width: 650px;
+`;
+
+type Props = {
+  isModalVisible :boolean;
+  onCloseModal :() => void;
+  title :string;
+  description :string;
+  rruleSet :string;
+  questionnaireEKID :UUID;
```
giuliacampana: please alphabetize :)

src/containers/questionnaires/components/QuestionnaireListItem.js
```diff
@@ -0,0 +1,111 @@
...
+
+const { SHOW_DETAILS, TOGGLE_STATUS } = LIST_ITEM_ACTIONS;
+const { NEUTRAL, PURPLE } = Colors;
+
+const Description = styled.div`
+  h3 {
+    padding: 0;
+    margin: 0 0 5px 0;
+    font-weight: 500;
+    color: ${NEUTRAL.N800};
```
giuliacampana: please alphabetize :)

src/containers/questionnaires/components/QuestionnaireListItem.js
```diff
@@ -0,0 +1,111 @@
...
+  }
+
+  p {
+    padding: 0;
+    margin: 0;
+    font-size: 15px;
+    color: ${NEUTRAL.N600};
+    white-space: nowrap;
+    overflow: hidden;
+    text-overflow: ellipsis;
```
giuliacampana: please alphabetize :)

src/containers/questionnaires/components/QuestionnaireListItem.js
```diff
@@ -0,0 +1,111 @@
...
+  :hover {
+    cursor: pointer;
+  }
+`;
+
+const IconGrid = styled.div`
+  display: flex;
+  grid-gap: 25px;
+  justify-content: space-between;
+  align-items: center;
```
giuliacampana: please alphabetize :)

src/containers/questionnaires/components/QuestionnaireListItem.js
```diff
@@ -0,0 +1,111 @@
...
+`;
+
+const ListItemWrapper = styled.div`
+  display: grid;
+  grid-template-columns: 1fr auto 60px;
+  grid-gap: 40px;
+  border-bottom: 1px solid #eee;
+  align-items: center;
+  padding: 20px 30px;
```
giuliacampana: please alphabetize :)

src/containers/questionnaires/components/QuestionnairePreview.js
```diff
@@ -0,0 +1,76 @@
...
+  font-weight: 400;
+  margin: 0;
+  padding: 0;
+`;
+
+const Description = styled.h6`
+  font-size: 16px;
+  font-weight: 400;
+  padding: 0;
+  margin: 5px 0 20px 0;
```
giuliacampana: please alphabetize :)

src/core/sagas/Sagas.js
```diff
@@ -12,6 +12,7 @@ import * as EDMSagas from '../edm/EDMSagas';
 import * as RoutingSagas from '../router/RoutingSagas';
 import * as StudiesSagas from '../../containers/studies/StudiesSagas';
 import * as SurveySagas from '../../containers/survey/SurveySagas';
+// import * as QuestionnairesSagas from '../../containers/questionnaires/QuestionnairesSagas';
```
giuliacampana: looks like you can delete this comment.

src/containers/questionnaires/components/NewQuestionnaireConfirmation.js
anzioka: No, it's not necessary. This component is rendered after paging through a bunch of required form inputs, from which the summary is computed. So the values rendered here are based on the form data which has already gone through  rjsf validation.

src/containers/questionnaire/QuestionnaireSagas.js
anzioka: Following our conversation with @jokedurnez, we are going with HARD delete for now.

solomontang (approved):

UnsungHero97 (approved):

PR #160, opened 2020-06-30
upgrading lattice dependencies

UnsungHero97:


src/utils/NotificationsUtils.js
```diff
@@ -1,16 +1,12 @@
...
-  const entityNamePrefix = 'chronicle_notifications_';
-  return `${entityNamePrefix}${studyId}`;
-};
+const getNotificationsEntitySetName = (studyId :UUID) => `${NOTIFICATIONS_PREFIX}${studyId}`;
+const getPartOfAssociationEntitySetName = (studyId :UUID) => `${PART_OF_PREFIX}${studyId}`;
 
 export {
   getNotificationsEntitySetName,
-  getPartOfAssociationEntitySetName
+  getPartOfAssociationEntitySetName,
```
anzioka: This file can be deleted since the exported functions are not used anywhere in the code base

src/utils/testing/InvalidParams.js
```diff
@@ -0,0 +1,26 @@
...
+  ' ',                            // 14
+  new String(),                   // 15
+  'invalid_special_string_value', // 16
+  /regex/,                        // 17
+];
+/* eslint-enable */
+
+export {
+  INVALID_PARAMS,
+};
```
anzioka, line 26: where is this used?

anzioka (changes requested):

src/utils/testing/InvalidParams.js
UnsungHero97: It used to be used in `TestUtils`, but it seems it's no longer being used after my cleanup. I'd like to keep it around since it will be necessary in the future (assuming more tests are written), and because other projects follow the same pattern.

anzioka (approved):

PR #401, opened 2020-11-24
Time use diary bug fix

anzioka:
PR:
This PR fixes an issue where the survey form entered an error state when a user, after filling out the survey, modifies the day start/end times in such a way as to cause certain data sections in PagedData to be out of range, thus interfering with validate on submit.

### DIFF
--- a/src/containers/tud/components/QuestionnaireForm.js
+++ b/src/containers/tud/components/QuestionnaireForm.js
@@ -80,10 +80,12 @@ const removeExtraData = (formRef :Object, pagedData, page :number) => {
     if (!currEndDateTime.equals(dayEndTime)) {
       return;
     }
+
     const pages = Object.keys(formData)
-      .map((key) => parsePageSectionKey(key))
-      .map((parsedResult) => parsedResult.page)
-      .map((pageStr) => Number(pageStr));
+      .map((key) => {
+        const parsed = parsePageSectionKey(key);
+        return Number(parsed.page);
+      });
 
     let toRemoveStartIndex = page + 1;
     // if the next page has night activity page data, skip
--- a/src/containers/tud/components/QuestionnaireForm.js
+++ b/src/containers/tud/components/QuestionnaireForm.js
@@ -3,10 +3,10 @@
 import React from 'react';
 
 import styled from 'styled-components';
-import { getIn, setIn, merge } from 'immutable';
+import { getIn, merge, setIn } from 'immutable';
 import { DataProcessingUtils, Form } from 'lattice-fabricate';
 import { Button } from 'lattice-ui-kit';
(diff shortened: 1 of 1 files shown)


src/containers/tud/components/QuestionnaireForm.js
```diff
@@ -58,6 +58,50 @@ const ButtonRow = styled.div`
...
+  const dayEndTime :?DateTime = selectTimeByPageAndKey(DAY_SPAN_PAGE, DAY_END_TIME, pagedData);
+  const formData :Object = formRef?.current?.state?.formData || {};
+
+  const currEndTime = formData[psk]?.[ACTIVITY_END_TIME];
+
+  if (dayEndTime && currEndTime) {
+    const currEndDateTime :DateTime = DateTime.fromSQL(currEndTime);
+    if (!currEndDateTime.equals(dayEndTime)) {
+      return;
+    }
```
UnsungHero97, line 82: Wait, don't you want to proceed if they are not equal?

> and later changes the day end time value

So if they change the value, whatever is in `formData` would be different, no? In which case, since the values are different, wouldn't you want to proceed?

src/containers/tud/components/QuestionnaireForm.js
```diff
@@ -58,6 +58,50 @@ const ButtonRow = styled.div`
...
+    const pages = Object.keys(formData)
+      .map((key) => parsePageSectionKey(key))
+      .map((parsedResult) => parsedResult.page)
+      .map((pageStr) => Number(pageStr));
```
UnsungHero97: You can do this in 1 `.map()`, why did you go with 3 `.map()`?

```
const pages = Object.keys(formData)
  .map((key :string) => {
    const psk = parsePageSectionKey(key);
    return Number(psk.page);
  });
```

src/containers/tud/components/QuestionnaireForm.js
anzioka: we only need to do this "forced mutation before submit" workaround when the user indicates that they are done filling the daytime part of the survey (that is when dayEndTime === endTime). Therefore, if they are not equal we don't need to mutate anything; we just execute Paged's validateAndSumit as usual.

src/containers/tud/components/QuestionnaireForm.js
anzioka: the suggested has arguably better performance, so will switch.

src/containers/tud/components/QuestionnaireForm.js
UnsungHero97: Got it. Thanks!

UnsungHero97 (approved):

giuliacampana (approved):

PR #416, opened 2020-12-04
2020-12-10 release

anzioka:


UnsungHero97 (changes requested):
Blocking until we communicate with customers about legacy.

solomontang (approved):

UnsungHero97 (approved):

PR #618, opened 2021-04-19
Delete participants legacy

anzioka:


solomontang (dismissed):

bergieboy (dismissed):

src/containers/studies/StudiesSagas.js
```diff
@@ -269,17 +266,32 @@ function* deleteStudyParticipantWorker(action :SequenceAction) :Generator<*, *,
 
     const { studyId, participantEntityKeyId, participantId } = action.value;
 
-    const { response, timeout } = yield race({
-      response: call(ChronicleApi.deleteStudyParticipant, participantId, studyId),
-      timeout: delay(1000 * 10) // 10 seconds
-    });
-    if (response && response.error) throw response.error;
+    yield call(ChronicleApi.deleteStudyParticipant, participantId, studyId);
```
UnsungHero97, line 269: You still need to check the response...

```
const response = yield call(ChronicleApi.deleteStudyParticipant, participantId, studyId);
if (response.error) throw response.error;
```

src/containers/studies/StudiesSagas.js
```diff
@@ -269,17 +266,32 @@ function* deleteStudyParticipantWorker(action :SequenceAction) :Generator<*, *,
...
+    const response = yield call(updateEntityDataWorker, updateEntityData({
+      entities: {
+        [participatedInEKID]: {
+          [statusPTID]: [DELETE],
+        }
+      },
+      entitySetId: participatedInESID,
+      updateType: UpdateTypes.PartialReplace
+    }));
+    if (response.error) throw response.error;
```
UnsungHero97, line 288: This is an edge case ... if the above delete succeeds but this update fails, you won't get another chance to make this update. Will things break if this happens?

UnsungHero97 (changes requested):

src/containers/studies/StudiesSagas.js
anzioka: I don't think we need to expressly check for errors since they will still be caught in the `catch` block

src/containers/studies/StudiesSagas.js
anzioka: Great question. An alternative approach might be to reverse the calls (make the update, then call delete), but that approach also has a similar loop hole: if the update succeeds and the delete fails, then we will have erroneously marked the study as deleted, and the user won't get another chance to actually delete

The current approach of updating after a successful delete job scheduling is arguably better, but not perfect. If the update fails, the user might still view the study but it will eventually disappear from the UI once the delete job is finished.

src/containers/studies/StudiesSagas.js
UnsungHero97: Right! This is not `lattice-sagas`.

UnsungHero97 (dismissed):

solomontang (approved):

UnsungHero97 (approved):

PR #706, opened 2021-06-11
Feature/apps 2933 download summarized tud data

anzioka:
PR:
- Add functionality to download summarized data:


src/containers/tud/constants/SummarizedDataVariables.js
```diff
@@ -0,0 +1,49 @@
+// flow
+
+import { OrderedSet } from 'immutable';
+
+// These are the column headers in the csv summary data
+// TODO: verify that this is the expected order
```
solomontang: This seems like one of those TODOs that needs to be done before this PR goes live. @anzioka

solomontang (dismissed):

src/containers/tud/constants/SummarizedDataVariables.js
anzioka: This file is no longer needed. column headers are generated from `ol.variable` values in the entity set data

src/containers/tud/TimeUseDiaryReducer.js
```diff
@@ -71,32 +70,26 @@ export default function timeUseDiaryReducer(state :Map = INITIAL_STATE, action :
...
-      return downloadTudResponses.reducer(state, action, {
-        REQUEST: () => state
-          .setIn([DOWNLOAD_TUD_RESPONSES, REQUEST_STATE, date, dataType], RequestStates.PENDING)
-          .setIn([DOWNLOAD_TUD_RESPONSES, action.id], action),
+      return downloadTudData.reducer(state, action, {
+        REQUEST: () => {
+          if (!date) {
+            return state.setIn([DOWNLOAD_ALL_DATA, REQUEST_STATE], RequestStates.PENDING);
+          }
+          return state.setIn([DOWNLOAD_TUD_DATA, REQUEST_STATE, date, dataType], RequestStates.FAILURE);
```
UnsungHero97: Since you are merging 2 reducers into 1, you should update both `REQUEST_STATE` values to avoid getting into a weird state. 

Additionally, while this is straightforward logic, I don't think it belongs here. I think `REQUEST: () => {}` should always set `RequestStates.PENDING` and the saga should be the one responsible for deciding when to call `.failure()` and when to call `.success()`.

src/containers/tud/TimeUseDiaryReducer.js
```diff
@@ -71,32 +70,26 @@ export default function timeUseDiaryReducer(state :Map = INITIAL_STATE, action :
...
-    case downloadAllData.case(action.type): {
-      return downloadAllData.reducer(state, action, {
-        REQUEST: () => state
-          .setIn([DOWNLOAD_ALL_DATA, REQUEST_STATE], RequestStates.PENDING)
-          .setIn([DOWNLOAD_ALL_DATA, action.id], action),
-        FAILURE: () => state
-          .setIn([DOWNLOAD_ALL_DATA, REQUEST_STATE], RequestStates.FAILURE),
-        SUCCESS: () => state
-          .setIn([DOWNLOAD_ALL_DATA, REQUEST_STATE], RequestStates.SUCCESS),
+          .setIn([DOWNLOAD_TUD_DATA, REQUEST_STATE, date, dataType], RequestStates.FAILURE),
```
UnsungHero97: `DOWNLOAD_ALL_DATA` update is missing.

src/containers/tud/TimeUseDiaryReducer.js
```diff
@@ -71,32 +70,26 @@ export default function timeUseDiaryReducer(state :Map = INITIAL_STATE, action :
...
+        SUCCESS: () => {
+          if (!date) {
+            return state.setIn([DOWNLOAD_ALL_DATA, REQUEST_STATE], RequestStates.SUCCESS);
+          }
+          return state
+            .setIn([DOWNLOAD_TUD_DATA, REQUEST_STATE, date, dataType], RequestStates.SUCCESS);
+        },
         FINALLY: () => state
-          .deleteIn([DOWNLOAD_ALL_DATA, action.id])
+          .deleteIn([DOWNLOAD_TUD_DATA, action.id])
```
UnsungHero97: `DOWNLOAD_ALL_DATA` update is missing.

src/containers/tud/TimeUseDiaryReducer.js
```diff
@@ -71,32 +70,26 @@ export default function timeUseDiaryReducer(state :Map = INITIAL_STATE, action :
...
-          .setIn([DOWNLOAD_ALL_DATA, REQUEST_STATE], RequestStates.FAILURE),
-        SUCCESS: () => state
-          .setIn([DOWNLOAD_ALL_DATA, REQUEST_STATE], RequestStates.SUCCESS),
+          .setIn([DOWNLOAD_TUD_DATA, REQUEST_STATE, date, dataType], RequestStates.FAILURE),
+        SUCCESS: () => {
+          if (!date) {
+            return state.setIn([DOWNLOAD_ALL_DATA, REQUEST_STATE], RequestStates.SUCCESS);
+          }
+          return state
+            .setIn([DOWNLOAD_TUD_DATA, REQUEST_STATE, date, dataType], RequestStates.SUCCESS);
```
UnsungHero97: Same comment.

src/containers/tud/TimeUseDiarySagas.js
```diff
@@ -213,17 +220,67 @@ function* getSubmissionsByDateWatcher() :Saga<*> {
...
+          const value = getPropertyValue(neighborDetails, [VALUES_FQN, 0]);
+          mutator.setIn([submissionId, variable], value);
+          csvHeaders.add(variable);
+        });
+      });
+    });
+
+    exportSummarizedDataToCsvFile(summaryData, submissionMetadata, csvHeaders.asImmutable(), outputFilename);
+
+    workerResponse = { data: null };
```
UnsungHero97: Setting it to `null` seems like an easy way to introduce a bug in the future.

```suggestion
    workerResponse = { data: {} };
```

src/containers/tud/TimeUseDiarySagas.js
```diff
@@ -213,17 +220,67 @@ function* getSubmissionsByDateWatcher() :Saga<*> {
...
+    const response = yield call(
+      searchEntityNeighborsWithFilterWorker,
+      searchEntityNeighborsWithFilter({
+        entitySetId: submissionESID,
+        filter,
+      })
+    );
+    if (response.error) throw response.error;
+
+    const csvHeaders = OrderedSet().asMutable();
```
UnsungHero97, line 256: No change required, just a thought. I'm not sure if the search response will always have the same order of neighbors. It seems possible that you'll get a different order of headers over time, which could potentially break any scripts researchers create that depend on the order of the columns.

src/containers/tud/components/SummaryListComponent.js
```diff
@@ -27,6 +27,12 @@ const Wrapper = styled.div`
   }
 `;
 
+const ButtonGrid = styled.div`
+  display: grid;
+  grid-template-columns: auto auto auto;
+  grid-gap: 20px;
+`;
```
UnsungHero97: You've been using `Grid` from MUI, right? Why not use it here as well?

src/containers/tud/utils/index.js
```diff
@@ -518,6 +519,38 @@ function writeToCsvFile(
   FS.saveAs(blob, outputFileName);
 }
 
+function exportSummarizedDataToCsvFile(
```
UnsungHero97, line 522: Why didn't you do this in the server?

UnsungHero97 (changes requested):

src/containers/tud/TimeUseDiaryReducer.js
anzioka: Line 80 was a typo. Updated to `RequestStates.PENDING`

anzioka: > Since you are merging 2 reducers into 1, you should update both `REQUEST_STATE` values to avoid getting into a weird state.
> 
Only one of the `REQUEST_STATE` values can be updated. For example, if no date value is passed then it wouldn't make sense to update the `[DOWNLOAD_TUD_DATA, REQUEST_STATE, date, dataType]` path

src/containers/tud/utils/index.js
anzioka: Not computation expensive, should be safe to do from browser

src/containers/tud/TimeUseDiaryReducer.js
```diff
@@ -71,32 +70,35 @@ export default function timeUseDiaryReducer(state :Map = INITIAL_STATE, action :
       });
     }
 
-    case downloadTudResponses.case(action.type): {
+    case downloadTudData.case(action.type): {
```
UnsungHero97: I don't understand the difference between having 2 dedicated reducers and a single merged reducer with an if-check in every case ... just seems like you're mixing logic with no clear benefit.

UnsungHero97 (dismissed):

UnsungHero97 (approved):

solomontang (approved):

UnsungHero97:
Waiting on app config changes.

PR #743, opened 2021-06-28
disable delete participant data if user not authorized

anzioka:


src/core/permissions/PermissionsSagas.js
```diff
@@ -215,9 +219,46 @@ function* getStudyAuthorizationsWatcher() :Generator<*, *, *> {
...
+
+    const allEntityKeyIds = coreEntitySetIds
+      .merge(dataCollectionEntitySetids)
+      .merge(surveyEntitySetIds)
+      .valueSeq()
+      .toJS();
+
+    const acls = allEntityKeyIds.map((ekid) => [ekid]);
+
+    const response = yield call(getAclsWorker, getAcls(acls));
```
UnsungHero97, line 239: No change required. Since we don't need the response for this request, perhaps we should add a new endpoint to `PermissionsApi` that only does the `ensurePermission` check. That being said, an endpoint like this might be a problem from a security perspective 🤔 .

src/core/permissions/PermissionsReducer.js
```diff
@@ -36,6 +43,19 @@ export default function permissionsReducer(state :Map = INITIAL_STATE, action :O
       });
     }
 
+    case getDeletePermission.case(action.type): {
+      return getDeletePermission.reducer(state, action, {
+        REQUEST: () => state.setIn([GET_DELETE_PERMISSION, REQUEST_STATE], RequestStates.PENDING),
+        FAILURE: () => state
+          .setIn([GET_DELETE_PERMISSION, REQUEST_STATE], RequestStates.PENDING)
```
UnsungHero97:
```suggestion
          .setIn([GET_DELETE_PERMISSION, REQUEST_STATE], RequestStates.FAILURE)
```

src/core/permissions/PermissionsSagas.js
```diff
@@ -215,9 +219,46 @@ function* getStudyAuthorizationsWatcher() :Generator<*, *, *> {
...
 }
 
+function* getDeletePermissionWorker(action :SequenceAction) :Saga<*> {
+  try {
+    yield put(getDeletePermission.request(action.id));
+
+    // to delete study/participant data, current user should have OWNER on all the entity set ids
+    const coreEntitySetIds = yield select(selectEntitySetsByModule(AppModules.CHRONICLE_CORE));
+    const dataCollectionEntitySetids = yield select(selectEntitySetsByModule(AppModules.DATA_COLLECTION));
+    const surveyEntitySetIds :Map = yield select(selectEntitySetsByModule(AppModules.QUESTIONNAIRES));
```
UnsungHero97, line 229: What about the participants entity set?

UnsungHero97 (changes requested):

UnsungHero97 (approved):

solomontang (approved):

PR #976, opened 2021-11-11
fix computing last/first android data

anzioka:
PR:
- Fixes a bug where "First Android Data" value for a participant is regarded as the first value in a list of date values mapped by ol.datetimestart. The value is correctly computed by taking the minimum of the values in the array


anzioka:
duplicate

PR #1048, opened 2021-12-22
Task/apps 3306/hourly usage survey

anzioka:
Customized app usage survey
intro
step1
step2
step3
step4
confirm_modal
modal2

success
modal1

desktop


UnsungHero97 (dismissed):

src/containers/survey/HourlyAppUsageSurvey.js
```diff
@@ -0,0 +1,241 @@
...
+        selected.delete(appName);
+      }
+      else {
+        selected.add(appName);
+      }
+
+      if (childOnly) {
+        return {
+          ...state,
+          childOnlyApps: selected
```
solomontang, line 60: when returning the new state, shouldn't the properties declared with `.asMutable()` be returned with `.asImmutable()`? Otherwise it seems like they're left in their .asMutable() state which defeats the purpose of using Immutable collections.

src/containers/survey/HourlyAppUsageSurvey.js
```diff
@@ -0,0 +1,241 @@
...
+  useEffect(() => {
+    if (isSubmissionConfirmed) {
+      storeDispatch(submitSurvey({
+        submissionData: createSubmissionData(),
+        organizationId,
+        participantId,
+        studyId
+      }));
+    }
+  }, [isSubmissionConfirmed]);
```
solomontang: Please do not ignore the React Hook useEffect dependency warning. Looks like you can just place createSubmissionData logic inside the hook.

https://reactjs.org/docs/hooks-faq.html#is-it-safe-to-omit-functions-from-the-list-of-dependencies

src/containers/survey/components/ConfirmSurveySubmissionModal.js
```diff
@@ -0,0 +1,31 @@
...
+import {
+  // $FlowFixMe
+  Box,
+  Modal
+} from 'lattice-ui-kit';
+
+import HourlySurveyDispatch, { ACTIONS } from './HourlySurveyDispatch';
+
+const ConfirmSurveySubmissionModal = () => {
+  const dispatch = useContext(HourlySurveyDispatch);
```
solomontang, line 13: very cool

solomontang (changes requested):

src/containers/survey/HourlyAppUsageSurvey.js
anzioka: No, they need to returned as mutable. Changing to immutable will introduce a bug. The first mutation will work fine, but subsequent mutations won't be possible.

src/containers/survey/HourlyAppUsageSurvey.js
```diff
@@ -173,21 +173,21 @@ const HourlyAppUsageSurvey = (props :Props) => {
     isInstructionsModalVisible
   } = state;
 
-  const createSubmissionData = () => {
+  useEffect(() => {
+    const createSubmissionData = () => {
```
UnsungHero97, line 177: No change required. An alternative would be to move this logic into the saga, where you can have more robust data processing all surrounded by a try-catch in case something is wrong. Passing a callback as a param is not ideal and we tend to avoid that pattern.

UnsungHero97 (approved):

solomontang (approved):

PR #1120, opened 2022-02-07
bugfix/APPS-3348 — enrollment link error page

giuliacampana:
Screen Shot 2022-02-11 at 11 41 51 AM


src/containers/app/AppContainer.js
```diff
@@ -80,6 +81,9 @@ const AppContainer = () => {
                   <DashboardContainer />
                 </Auth0AdminRoute>
               )} />
+          <Route path={Routes.ENROLLMENT_LINK}>
+            <EnrollmentLink />
+          </Route>
```
anzioka: This will not work if the user doesn't have a OL account (most likely to be the case) / doesn't have access to a study. `initializeApplicationRS === RequestStates.SUCCESS` requires both conditions to be satisfied. Please consider moving Route to `index.js`

giuliacampana: thank you, this is very helpful! i will update

src/containers/enrollment/EnrollmentLink.js
```diff
@@ -0,0 +1,42 @@
...
+  const enrollmentLink = `http://openlattice.com/login?${enrollmentLinkQueryParams}`;
+  return (
+    <>
+      <Card>
+        <CardHeader mode="danger">
+          Page Not Found
+        </CardHeader>
+        <CardSegment>
+          You have tried to open a mobile link in a web browser. Please install the Chronicle mobile application
+          and open the link on a mobile device.
```
anzioka: We need to update this message to apply to iOS users as well.

src/containers/enrollment/EnrollmentLink.js
```diff
@@ -0,0 +1,42 @@
...
+  Card,
+  CardHeader,
+  CardSegment,
+} from 'lattice-ui-kit';
+
+import copyToClipboard from '../../utils/copyToClipboard';
+
+const EnrollmentLink = () => {
+  const enrollmentLinkQueryParams = window.location.hash.split('?')[1];
+  const enrollmentLink = `http://openlattice.com/login?${enrollmentLinkQueryParams}`;
```
UnsungHero97: The link is missing `/chronicle` in the path.

```suggestion
  const enrollmentLink = `http://openlattice.com/chronicle/login?${enrollmentLinkQueryParams}`;
```

src/containers/enrollment/EnrollmentLink.js
```diff
@@ -0,0 +1,42 @@
...
+  Button,
+  Card,
+  CardHeader,
+  CardSegment,
+} from 'lattice-ui-kit';
+
+import copyToClipboard from '../../utils/copyToClipboard';
+
+const EnrollmentLink = () => {
+  const enrollmentLinkQueryParams = window.location.hash.split('?')[1];
```
UnsungHero97: I don't think this works. Here's an example of a valid link:

```
https://openlattice.com/chronicle/login?organizationId=...&studyId=...&participantId=...
```

There's no hash for this link so `window.location.hash` will return an empty string. Furthermore, to get the query string, you can just use the `search` property:

https://developer.mozilla.org/en-US/docs/Web/API/Location/search

```
const qs = window.location.search;
// qs = "?organizationId=...&studyId=...&participantId=..."
```

UnsungHero97 (changes requested):

src/containers/enrollment/EnrollmentLink.js
giuliacampana: ohh hmm i was thinking this message was generic enough to apply to either iOS or android? what do you think?

anzioka: ah never mind. Thought I saw "Android" in the text

src/containers/enrollment/EnrollmentLink.js
giuliacampana: oops, thank you!

src/containers/enrollment/EnrollmentLink.js
UnsungHero97: I think the wording needs to change a bit. I think it would be clearer to refer to it as the "enrollment link". Additionally, the behavior is the same whether you open it on mobile/desktop, i.e. you still get a 404. How about something like this ...

> You have tried to open the Chronicle enrollment link in a web browser. Please install the Chronicle mobile application and open the enrollment link with it.

I don't like the "with it" at the end there, just couldn't think of a better suggestion.

anzioka: how about ... `and open the link on your device.`

giuliacampana: Oh good point. Thanks! I changed it to this—does this make sense?

`You have tried to open the Chronicle enrollment link in a web browser. Please install the
          Chronicle mobile application and try opening the link again.`

src/containers/enrollment/EnrollmentLink.js
giuliacampana: I understand what you're saying and also initially thought I should use the `search` property to get the query string...but I think I've confused myself as to what I'm trying to accomplish here. Whether logged in or not, navigating to the URL will always add a hash to end of the URL, so the correct page is not rendered. (When logged in, it just redirects to the studies page, which is expected because the AppContainer has a `Redirect` element pointing to that route, and when not logged in, it redirects to `/#/login/`, the actual login page.) Of course, when there is hash included after `/chronicle/`, the page we expect is rendered just fine:

```
https://openlattice.com/chronicle/#/login?organizationId=...&studyId=...&participantId=...
```

Also... I'm now not sure the query params need to be included at all in the URL. Because I'm not sure it makes sense to even include a button to copy the enrollment link, which is the only place the params are being used. I'm not sure why the user would be copying the link at all, if they've received it via email—I would imagine they'd install the app and then try clicking on the URL again? I could be misunderstanding.
But in that case, I think having the enrollment link route to `https://openlattice.com/chronicle/login/enrollment/` would be sufficient. How does this work in terms of making AWS changes?

src/index.js
```diff
@@ -77,49 +78,71 @@ const GlobalStyle = createGlobalStyle`
...
-                <GlobalStyle />
-              </>
-            </StylesProvider>
-          </MuiPickersUtilsProvider>
-        </ThemeProvider>
-      </Provider>
-    </Suspense>,
-    APP_ROOT_NODE
-  );
+if (window.location.pathname === '/chronicle/login/enrollment') {
```
UnsungHero97:
```suggestion
if (window.location.pathname === '/chronicle/login') {
```

UnsungHero97 (changes requested):

src/index.js
giuliacampana: oops thank you!

UnsungHero97 (approved):

anzioka (approved):

PR #2030, opened 2023-12-30
add missing spanish translations

UnsungHero97:


geekbeast (approved):
LGTM will share with Bolim before merging

